PR #66, opened 2024-09-20
Btfhub arm64

haesbaert:
Btfhub table for arm64
While here, adjust a couple of things:
	- remove an unused offset
	- add missing fedora 30/31 to amd64

Preparation for upcoming changes where we try to match the best kernel, and thus
actually use the table.

### DIFF
--- a/genbtf.sh
+++ b/genbtf.sh
@@ -20,8 +20,9 @@ if [ $# -ne 1 ]; then
    usage
 fi
 
-# amd64 only for now
-Srcs="
+Btfhub_path="$1"
+
+Srcs_amd64="
 amzn/2/x86_64
 amzn/2018/x86_64
 centos/7/x86_64
@@ -35,6 +36,8 @@ fedora/26/x86_64
 fedora/27/x86_64
 fedora/28/x86_64
 fedora/29/x86_64
+fedora/30/x86_64
+fedora/31/x86_64
 ol/7/x86_64
 ol/8/x86_64
 rhel/7/x86_64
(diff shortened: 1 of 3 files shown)


genbtf.sh
```diff
@@ -64,48 +97,75 @@ cat <<EOF
...
+		Arch=amd64
+		Srcs="$Srcs_amd64"
+	elif [ $1 = arm64 ]; then
+		Arch=aarch64
+		Srcs="$Srcs_arm64"
+	else
+		die "bad arch"
+	fi
+
+	printf "#ifdef __%s__\n\n" $Arch
```
nicholasberlin, line 115: Do you need this? Or line 140?

genbtf.sh
```diff
@@ -64,48 +97,75 @@ cat <<EOF
...
+do_arch arm64
 
 cat <<EOF
 struct quark_btf *all_btfs[] = {
 EOF
 
-for name in "${Good[@]}"; do
-	printf "\t&%s,\n" "$name"
-done
+	printf "\n#ifdef __amd64__\n"
```
nicholasberlin, line 151: Weird spacing in these 10 lines.

genbtf.sh
haesbaert: I do, this is purely to save space, basically an amd64 build will only compile the amd64 entries, and an arm64 only the arm64 one.

genbtf.sh
haesbaert: whoa, good catch, no clue how I didn't spot it.

haesbaert: arghh ah no, wait, sorry, lol.
Now I remembered, this is intentional, it's to look like it's inside the "all_btfs" block.
Does it make sense?

nicholasberlin: Oh, I see. I wouldn't have figured that out, since the printf's still need to have `\t`s in them.

nicholasberlin (approved):

PR #75, opened 2024-10-08
Logging functions, issue #2

haesbaert:
Logging functions

follows the text of the first commit, each commit details itself.

--

Implement logging functions. Issue #2
First of 3 commits.

This implements qwarn, qwarnx, qdebug and qdebugx, they behave a bit like
warn(3).

Conditionally only print depending on quark_verbose, the output has both
`program_invocation_short_name` and the `function:line` prepended. Example:

```
quark-mon: quark_btf_open2:316: dotname=task_struct.pids failed
```

The dance around allocating a new fmt in nfmt is to attempt the whole output in
one syscall, preventing interleaved output.

I chose only 3 levels of printing, since in the normal case it should be
"none/silent", as libraries shouldn't print. Then the next would be either info
or warning, but having both made no sense imho.

So it's basically:
  - SILENT, normal behaviour
  - WARNING, show me some stuff so I can see what's going wrong
  - DEBUG, print everything no one needs

`quark_verbose` can be influenced by passing `-v` multiple times in quark-mon(8)
and quark-btf(8).

Output example in trying to start quark-mon(8) without priviledges:
```
eru:quark: ./quark-mon -sk
quark-mon: quark_queue_open: Permission denied
```

with WARNING:
```
eru:quark: ./quark-mon -skv
quark-mon: open_tracing:703: open: events/sched/sched_process_exec/format:
Permission denied
quark-mon: open_tracing:695: open: /sys/kernel/debug/tracing: Permission denied
quark-mon: quark_queue_open:1734: all backends failed
quark-mon: quark_queue_open: Permission denied
```

### DIFF
--- a/qutil.c
+++ b/qutil.c
@@ -334,7 +334,7 @@ qlog_func(int pri, int do_errno, const char *func, int lineno,
 		    program_invocation_short_name, func, lineno);
 		vfprintf(stderr, fmt, ap);
 		if (do_errno)
-			fprintf(stderr, "%s:", strerror(saved_errno));
+			fprintf(stderr, ": %s", strerror(saved_errno));
 		fprintf(stderr, "\n");
 	} else {
 		if (do_errno) {
--- a/qutil.c
+++ b/qutil.c
@@ -6,6 +6,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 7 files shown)


qutil.c
```diff
@@ -310,3 +311,39 @@ args_free(struct args *args) {
...
+
+	va_start(ap, fmt);
+
+	/* best effort in out of mem situations */
+	if (asprintf(&nfmt, "%s:%d: %s", func, lineno, fmt) == -1) {
+		fprintf(stderr, "%s: %s:%d: ",
+		    program_invocation_short_name, func, lineno);
+		vfprintf(stderr, fmt, ap);
+		if (do_errno)
+			fprintf(stderr, "%s:", strerror(saved_errno));
```
nicholasberlin, line 337: why the trailing `:`?

nicholasberlin (approved):

qutil.c
haesbaert: Cause I'm a tool :rofl:, I'll fix it 
You have good eyes!

PR #78, opened 2024-10-12
docs: Fix typo in index.html quickstart instructions

palash25:


haesbaert:
Heya thanks for the fix!

index.html is auto-generated from quark.7 (which also generates quark.7.html), we commit them so that it can be displayed in github.io.

haesbaert (approved):

PR #79, opened 2024-10-14
Add basic versioning scheme

haesbaert:
QUARK_VERSION should be shared by libquark, quark-btf, quark-mon and also elastic/go-quark.

A release has the form "MAJOR.MINOR", we break ABI *freely*, so a MINOR can break ABI. MAJOR denotes a major milestone, or some cool stuff that was added.

Quark is not meant to be distributed as a shared object, therefore it makes no sense for us to be careful with ABI breakages, users should be building quark together with their application. We should document them in a CHANGES file for every release.

The suffix "a" is added to the version we are working on, so when we're about to release: remove the suffix; release; bump with suffix.

tip for review: ignore the html part as it's autogenerated.

cc @mjwolf

### DIFF
--- a/docs/quark-btf.8.html
+++ b/docs/quark-btf.8.html
@@ -45,6 +45,13 @@
       <var class="Ar">btf_file name version</var></td>
   </tr>
 </table>
+<br/>
+<table class="Nm">
+  <tr>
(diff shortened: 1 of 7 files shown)


haesbaert:
ignore CI as it's pending on another PR

nfritts (approved):

PR #80, opened 2024-10-14
Zap Go, moved to https://github.com/elastic/go-quark

haesbaert:
We moved the Go files to another repository as we have to ship the archives in order to ease go distribution.

If we kept the files here, we would have to rebuild and commit them at every step of development which would bloat the repository size considerably.

In go-quark we will only update quark itself when we're about to release.

cc @mjwolf


haesbaert:
Ignore CI as it's pending on another commit.

nfritts (approved):

PR #84, opened 2024-10-18
Stash the active backend in quark_queue_stats{}

haesbaert:
The whole idea of "the user should not even know which backend is being used", worked a bit too much.

This was prompted by a bug in beats where seccomp was blocking a syscall needed for EBPF, and so it fell back to kprobe all the time and the user had no proper way to notice this.

Related to https://github.com/elastic/beats/pull/41297

Purposedly stashed this into stats to make it hard to retrieve from quark.c, if we start doing `if (qq->backend == QQ_EBPF)` we lost the war.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,0 +1,9 @@
+Quark Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~
+  This file documents breaking and/or other changes that might require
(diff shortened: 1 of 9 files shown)


nicholasberlin (approved):
LGTM

PR #89, opened 2024-10-25
Run quark-test in the CI

haesbaert:
Run quark-test from the centos7 build as a CI target, this is the very beginning
    and we run it only in the same machine that we make a build.

### DIFF
--- a/.buildkite/runtest.sh
+++ b/.buildkite/runtest.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -euo pipefail
+
+function download {
+	buildkite-agent artifact download "$1" "$2"
+}
+
+if [ -z "${BUILDKITE}" ]; then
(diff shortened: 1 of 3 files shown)


.buildkite/runtest.sh
```diff
@@ -0,0 +1,19 @@
...
+}
+
+if [ -z "${BUILDKITE}" ]; then
+	echo "This script doesn't appear to be running in buildkite" 1>&2
+	echo "refusing to continue" 1>&2
+	exit 1
+fi
+
+download quark-test .
+chmod +x quark-test
```
mjwolf, line 16: I'm not sure if this is needed, I think it will keep the permissions from the upload, but it doesn't hurt to have

haesbaert: Seems to be, I did a quick test and it failed without.

mjwolf (approved):

PR #91, opened 2024-10-27
Fix pgid/sid for RHEL8 and improve kprobe expressions

haesbaert:
RHEL8 has `pids` in signal_struct{} but has the old value for PIDTYPE_PGID.

This bug is _also_ in ebpf and in elastic/ebpf, probably since forever as it's pretty hard to detect. I'll put another PR for elastic/ebpf. This fixes kprobe only and ebpf will still fail with:
```
[haesbaert@rocky8 quark]$ sudo ./quark-test
t_probe @ ebpf: ok
t_probe @ kprobe: ok
t_fork_exec_exit @ ebpf: failed
quark-test: quark-test.c:198: t_fork_exec_exit: Assertion `(pid_t)qp->proc_pgid == getpgid(0)' failed.
t_fork_exec_exit @ kprobe: ok
failed tests 1
````

old kernels: https://elixir.bootlin.com/linux/v3.10.108/source/include/linux/pid.h#L69
newer kernels: https://elixir.bootlin.com/linux/v6.12-rc4/source/include/linux/pid_types.h#L10

There are basically two changes, the whole `pids` thing moved from `task_struct` to `signal_struct`, _and_ PIDTYPE_PGID and PIDTYPE_SID got bumped in newer kernels. The bug is because we always assumed both changes occurred together, then comes RHEL, which moved `pids` to `signal_struct` but didn't bump PIDTYPE_PGID.

The fix is teaching quark about enums@BTF and improving the expression parser to accept recursive arithmetics.

This also cleans up the parser a bit, I'm not proud of it, but it looks a bit better than before.

Discovered by running the upcoming "test-this-on-all-kernels" script.

### DIFF
--- a/kprobe_queue.c
+++ b/kprobe_queue.c
@@ -541,7 +541,7 @@ perf_event_to_raw(struct quark_queue *qq, struct perf_event *ev)
 		qq->stats.lost += ev->lost.lost;
 		break;
 	default:
-		qwarnx("unhandled type %d\n", ev->header.type);
+		qwarnx("unhandled type %d", ev->header.type);
 		return (NULL);
 		break;
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: kprobe_queue.c [hunk 1]
```diff
@@ -787,59 +787,128 @@ parse_data_offset(void)
...
+	}
+	if (strlcpy(left, start_left, TOKSZ) >= TOKSZ) {
+		qwarnx("left token overflow");
+		return (-1);
+	}
+	if (strlcpy(right, start_right, TOKSZ) >= TOKSZ) {
+		qwarnx("right token overflow");
+		return (-1);
+	}
+
```
nicholasberlin, line 854: Add length checks for left and right?

I think mistakes like `(+16)` or `(16+)` will result in hard to understand error messages later.

haesbaert: you mean is left or right is empty?

nicholasberlin (approved):

PR #100, opened 2024-11-12
Add kernel tests and improve testing infrastructure

haesbaert:
main commit:
```
    This adds targets `test` `test-kernel` and `test-all` aslong with the scripts
    supporting them `krun.sh` and `ktest-all.sh`.

    `krun.sh` can be used to run a quark utility in a specified kernel:

    $ ./krun.sh initramfs.gz kernel-images/amd64/linux-6.11.4-101.fc39.x86_64 quark-test -v

    `ktest-all.sh` is just a wrapper for running all kernels in kernel-images.

    Keeping the idea of not running scripts directly, these will mostly be used via
    `test-kernel`.
```
Each commit documents itself, and they won't be squashed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -414,6 +414,8 @@ clean-docs:
 	test-kernel
 
 .NOTPARALLEL:			\
+	clean			\
+	clean-all		\
 	centos7			\
 	centos7-image		\
 	centos7-shell		\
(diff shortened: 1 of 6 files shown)


haesbaert:
> LGTM, all three make commands worked for me once I enabled kvm in my VMs.

Thanks, I removed the kernel commit, I'll see if we can have a better place to stash them, I'm merging the rest.

PR #104, opened 2024-11-21
Increase proc_time_boot precision and make it immutable

haesbaert:
If we get proc_time_boot from /proc, the precision is lower than when we get from the kernel via a fork/exec/exit event. In /proc the unit is clock ticks, and usually hz=100. In the kernel the unit is actual nanoseconds.

First change is to increase the precision by using the remainder of the clock division, instead of just taking ticks/hz by face value.

Second change is to never mutate proc_boot_time after we settled on it, if we first learned it through /proc (the less precise one), we will keep that forever throughout the lifetime of the process. Before this change we were always updating, meaning we would overwrite the less precise with the more precise.

The problem with "upgrading" is that user code might rely on that value being immutable, most notably, proc_boot_time is needed to calculate process.entity_id in ECS, therefore it must not mutate, or user code would have to degrade precision, which is bug prone, it also makes debugging a nightmare if you want to search by proc_boot_time.

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -1002,7 +1002,16 @@ raw_event_process(struct quark_queue *qq, struct raw_event *src, struct
 		qp->proc_cap_effective = raw_task->cap_effective;
 		qp->proc_cap_bset = raw_task->cap_bset;
 		qp->proc_cap_ambient = raw_task->cap_ambient;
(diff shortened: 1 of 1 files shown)


nicholasberlin (approved):
LGTM

PR #109, opened 2024-12-01
Replace quark_queue_get_events(3) with get_event(). Fixes #107

### LINKED ISSUE #107: quark_process may contain "future" state
This bug was introduced in 9d38ecf767a02a080ae8ed7bccf66ecf2281b3ee

Basically when we call quark_queue_get_events(3) we update the internal quark_process{},
so if you get the following events:
```
[0] pid 7 exec foo
[1] pid 7 exec bar
```

event 0 will show up with filename `bar` instead of `foo`, it gets overwritten!
Bug can be worked around for now by never using more than one slot on quark_queue_get_events(3).

-- 

On the commit that introduced this I basically abolished the storage "per event", but silly me never accounted for this.
We can fix it easily by changing the interface to return only one event:
```
const struct quark_event *quark_queue_pop_event(struct quark_queue **)
```

Which """might""" impact Go a bit as it will now be one cgo call per event, which we can fix it back by giving back the old api, you give N slots and each event is copied out, but I doubt this would be better than the nocopy 1call=1event.

haesbaert:
quark_queue_get_events(3) was initially designed when I wanted to copy everything out of quark, which was changed in
9d38ecf767a02a080ae8ed7bccf66ecf2281b3ee to return pointers to internal state.

The problem is that since that change we were getting "future" state on events of the same pid, on a quark_queue_get_events() == 2, we could have:

[0] pid 7 exec foo
[1] pid 7 exec bar

event [0] will show up with filename "bar" instead of "foo", since they update the internal cache and return the pointer to it.

This diff kills quark_queue_get_events(3) and replaces with a single quark_queue_get_event().

With this, quark is able to produce 3.4 million events per second on my laptop, so I think we're fine. If I find that this affects Go too much, which my initial tests show it doesn't, I can bring quark_queue_get_events(3) back in a way that copies out everything, like the first version.


nicholasberlin (approved):
LGTM

Issue #74, opened 2024-10-08
Remove the remaining uses of err(3)


haesbaert:
There are some panic calls that maybe we can remove, unwrap the stack and kill the queue, better than killing the program, these are basically extra bug checking so they would only trigger if we screwed up. Still, this is a library, and libraries shouldn't exit.
```
btf.c:          err(1, "bit offset not multiple of 8");
quark.c:                err(1, "clock_gettime");
quark.c:                err(1, "we got consecutive collisions, this is a bug");
quark.c:                err(1, "collision on pidtime tree, this is a bug");
kprobe_queue.c:         errx(1, "getting an event larger than wrapped buf");
```

PR #111, opened 2025-01-14
Remove remaining cases of err(3). Issue #74

haesbaert:
- Bump retries on insert from 10 to 1000 before giving up.
- Put a timeout on drain_for_pid(), I used to manually test the changing paths, and it's good to have anyway.
- Check if clock_gettime() works in quark_queue_open(), maybe it gets EPERM due to seccomp and whatnot, sanitize return of now64(), program still survives. I just don't want to check now64() errors while they're impossible.

### DIFF
--- a/kprobe_queue.c
+++ b/kprobe_queue.c
@@ -1467,7 +1467,10 @@ kprobe_queue_populate(struct quark_queue *qq)
 			empty_rings = 0;
 			raw = perf_event_to_raw(qq, ev);
 			if (raw != NULL) {
-				raw_event_insert(qq, raw);
+				if (raw_event_insert(qq, raw) == -1) {
+					raw_event_free(raw);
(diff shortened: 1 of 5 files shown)


kprobe_queue.c
```diff
@@ -1467,7 +1467,10 @@ kprobe_queue_populate(struct quark_queue *qq)
 			empty_rings = 0;
 			raw = perf_event_to_raw(qq, ev);
 			if (raw != NULL) {
-				raw_event_insert(qq, raw);
+				if (raw_event_insert(qq, raw) == -1) {
+					raw_event_free(raw);
+					raw = NULL;
```
nicholasberlin, line 1472: good catch setting this to NULL.

nicholasberlin (approved):

PR #112, opened 2025-01-21
TCP socket events for EBPF

haesbaert:
This implements basic socket events for TCP, it tracks
connect(2)/accept(2)/close(2) on respective sockets as well as builds the
initial snapshot of existing sockets by parsing /proc.

Sockets are expressed in quark_socket{} and correlate with processes. Since a
socket might be in more than one process, two fields are provided:
 - pid_origin, the pid that established the connection.
 - pid_last_use, the last pid that affected it, if another process did the final
   close, pid_origin remains the same, but pid_last_use points to the last one.

We stash the lookup of quark_process_lookup(pid_origin) in the event process
pointer, which *can* be NULL, and the socket itself is stashed in the socket
member.

quark_event{} might need some adjustments in the future.

The EBPF probes are a bit subpar and they currently have two issues:
 - A connection that never sent a byte doesn't produce an exit event.
   elastic/ebpf#217
 - Exit events happen before TCP close is issued
   elastic/ebpf#216

I don't think this is good enough and will consider it experimental until both
issues are solved, first one has a fix on the way, second one likely involves
rewriting the whole probe.

Quark sockets don't produce snapshot events, because snapshot will be removed
and the user will have to use the iterator if desired.

In order to test this, pass -S to quark-mon(8).

Co-authored-by: Nicholas Berlin <56366649+nicholasberlin@users.noreply.github.com>

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -2,8 +2,15 @@
 /* Copyright (c) 2024 Elastic NV */
 
 #include <sys/epoll.h>
+#include <sys/socket.h>
+
+#include <netinet/in.h>
+#include <netinet/tcp.h>
+
+#include <arpa/inet.h>
 
 #include <ctype.h>
+#include <dirent.h>
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -24,6 +31,10 @@
 static int	raw_event_by_time_cmp(struct raw_event *, struct raw_event *);
 static int	raw_event_by_pidtime_cmp(struct raw_event *, struct raw_event *);
 static int	process_by_pid_cmp(struct quark_process *, struct quark_process *);
+static int	socket_by_src_dst_cmp(struct quark_socket *, struct quark_socket *);
+
+static void	process_cache_delete(struct quark_queue *, struct quark_process *);
+static void	socket_cache_delete(struct quark_queue *, struct quark_socket *);
 
 /* For debugging */
 int	quark_verbose;
@@ -32,6 +43,11 @@ RB_PROTOTYPE(process_by_pid, quark_process,
     entry_by_pid, process_by_pid_cmp);
 RB_GENERATE(process_by_pid, quark_process,
     entry_by_pid, process_by_pid_cmp);
+
+RB_PROTOTYPE(socket_by_src_dst, quark_socket,
+    entry_by_src_dst, socket_by_src_dst_cmp);
+RB_GENERATE(socket_by_src_dst, quark_socket,
+    entry_by_src_dst, socket_by_src_dst_cmp);
 
 RB_PROTOTYPE(raw_event_by_time, raw_event,
     entry_by_time, raw_event_by_time_cmp);
@@ -75,6 +91,7 @@ raw_event_alloc(int type)
 		qstr_init(&raw->exec_connector.task.cwd);
 		break;
 	case RAW_COMM:		/* nada */
+	case RAW_SOCK_CONN:	/* nada */
 		break;
 	default:
 		qwarnx("unhandled raw_type %d", raw->type);
@@ -105,6 +122,7 @@ raw_event_free(struct raw_event *raw)
 		qstr_free(&raw->exec_connector.task.cwd);
 		break;
 	case RAW_COMM:		/* nada */
+	case RAW_SOCK_CONN:	/* nada */
 		break;
 	default:
 		qwarnx("unhandled raw_type %d", raw->type);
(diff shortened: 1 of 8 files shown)


quark.c
```diff
@@ -408,43 +479,179 @@ process_cache_inherit(struct quark_queue *qq, struct quark_process *qp, int ppid
...
+	if (sockaddr_to_str(&qsk->remote, buf, len) == -1)
+		return (-1);
+
+	return (0);
+}
+#endif
+
+static struct quark_socket *
+socket_cache_get(struct quark_queue *qq,
+    struct quark_sockaddr *local, struct quark_sockaddr *remote, int alloc)
```
nicholasberlin, line 551: can local and remote be const?

quark.c
```diff
@@ -408,43 +479,179 @@ process_cache_inherit(struct quark_queue *qq, struct quark_process *qp, int ppid
...
+		return (r);
+
+	cmplen = a->local.af == AF_INET ? 4 : 16;
+	r = memcmp(&a->local.addr6, &b->local.addr6, cmplen);
+
+	return (r);
+}
+
+const struct quark_socket *
+quark_socket_lookup(struct quark_queue *qq,
```
nicholasberlin, line 632: I don't see this function being called. Enlighten me please.

nicholasberlin: Oh, I suppose these are used by library users.

quark.c
```diff
@@ -1329,7 +1588,102 @@ sproc_namespace(struct quark_process *qp, const char *path, u32 *dst, int dfd)
...
+		    qsk) != NULL) {
+			qwarnx("socket collision, this is a bug");
+			free(qsk);
+			continue;
+		}
+		printf("pid %d fd %s -> %s (inode=%lu, ss=%p)\n", pid,
+		    d->d_name, buf, inode, ss);
+	}
+
+	/* closedir() closes the backing `fdfd` */
```
nicholasberlin, line 1678: I was wondering that. Good comment.

quark.c
```diff
@@ -1329,7 +1588,102 @@ sproc_namespace(struct quark_process *qp, const char *path, u32 *dst, int dfd)
...
+		 * Maybe another process already references the same socket.
+		 */
+		bzero(&qsk_key, sizeof(qsk_key));
+		qsk_key.local = ss->socket.local;
+		qsk_key.remote = ss->socket.remote;
+		qsk = RB_FIND(socket_by_src_dst, &qq->socket_by_src_dst,
+		    &qsk_key);
+		if (qsk != NULL) {
+			/* XXX TODO WRITE AN EXAMPLE TO TRIGGER THIS */
+			printf("EXISTING !!! pid %d fd %s -> %s (inode=%lu, ss=%p)\n", pid,
```
nicholasberlin, line 1652: Is this an example of a process forking without closing the socket?

quark.c
```diff
@@ -1362,12 +1716,214 @@ sproc_pid(struct quark_queue *qq, int pid, int dfd)
...
+
+	/*
+	 * We're tracking the active side, the stack goes to ESTABLISHED when it
+	 * receives the SYN/ACK. We go to TCP_CLOSE_WAIT when we get the FIN but
+	 * didn't close().
+	 */
+	if (state != TCP_ESTABLISHED && state != TCP_CLOSE_WAIT)
+		return (0);
+
+#if 1
```
nicholasberlin, line 1798: still need this debugging?

quark.c
```diff
@@ -1756,6 +2344,7 @@ quark_queue_default_attr(struct quark_queue_attr *qa)
 	bzero(qa, sizeof(*qa));
 
 	qa->flags = QQ_ALL_BACKENDS;
+	qa->flags |= QQ_NET;	/* XXX REMOVE ME XXX */
```
nicholasberlin, line 2347: reminder

quark.c
```diff
@@ -1996,6 +2590,57 @@ quark_queue_get_snap_event(struct quark_queue *qq)
...
+	case CONNECTION_CLOSE:
+		if (qsk->close_time == 0)
+			qsk->close_time = raw->time;
+		gc_mark(qq, &qsk->gc, GC_SOCKET);
+		qev->events = QUARK_EV_CONNECTION_CLOSED;
+
+		break;
+	default:
+		qwarnx("invalid conn->conn %d\n", conn->conn);
+		return (NULL);
```
nicholasberlin, line 2635: if `qsk` was allocated by `socket_cache_get`, should it be freed here?

quark.c
haesbaert: Yep, or sending the socket via sendmsg() to another process, but basically any case where a socket is referenced/used by more than one process.

quark.c
haesbaert: gonna zap it

quark.c
haesbaert: Yep, this is more like a programming error, but we should be correct anyway, I'll have to change this function a bit, because there is a second issue:
Basically socket_cache_get() cannot do the "lookup or alloc", it needs to be split into two. When we get an ACCEPT/CONNECT, we must hit the alloc case, which might incur us evicting the old entry and creating a new one.
This can be the case if, for any reason, we missed the close event.

quark.c
```diff
@@ -408,43 +479,179 @@ process_cache_inherit(struct quark_queue *qq, struct quark_process *qp, int ppid
...
 	RB_REMOVE(process_by_pid, &qq->process_by_pid, qp);
-	if (qp->gc_time)
-		TAILQ_REMOVE(&qq->event_gc, qp, entry_gc);
+	gc_unlink(qq, gc);
 	free(qp);
 }
 
 static int
-process_cache_gc(struct quark_queue *qq)
+process_by_pid_cmp(struct quark_process *a, struct quark_process *b)
```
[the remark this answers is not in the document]
haesbaert: They could, but all the other _cmp are not, so I'd just keep the same.
Generally I'm more worried about constifying external calls

quark.c
haesbaert: Could, as it's not an external facing function, I wouldn't care much (most of the other code is node const)

quark.c
haesbaert: Yep!

quark.c
haesbaert: In the upcoming version, it doesn't have qsk allocated on that path anymore, so this is gone.

quark.c
```diff
@@ -408,43 +479,179 @@ process_cache_inherit(struct quark_queue *qq, struct quark_process *qp, int ppid
...
-		process_cache_delete(qq, qp);
-		n++;
+	return (0);
+}
+
+/*
+ * Socket stuff
+ */
+
+#if 0
```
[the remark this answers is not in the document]
haesbaert: zapped them

haesbaert:
Pushed the proposed final version.
Relevant changes:
 - Renamed connection to sock
 - Reworked socket allocation and insertion
 - Addressed the review comments
 - Updated the commit message and this PR.

nicholasberlin (approved):

PR #115, opened 2025-01-23
Pre net2

haesbaert:
A couple of harmless fixes and importing ebpf files from elastic/ebpf.

Preparation for the upcoming network events PR.

### DIFF
--- a/bpf_queue.c
+++ b/bpf_queue.c
@@ -164,6 +164,8 @@ ebpf_events_to_raw(struct ebpf_event_header *ev)
 		break;
 	case EBPF_EVENT_PROCESS_EXEC:
 		exec = (struct ebpf_process_exec_event *)ev;
+		if (exec->pids.tid != exec->pids.tgid)
+			goto bad;
 		if ((raw = raw_event_alloc(RAW_EXEC)) == NULL)
 			goto bad;
 		raw->pid = exec->pids.tid;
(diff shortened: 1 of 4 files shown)


bpf_queue.c
```diff
@@ -164,6 +164,8 @@ ebpf_events_to_raw(struct ebpf_event_header *ev)
 		break;
 	case EBPF_EVENT_PROCESS_EXEC:
 		exec = (struct ebpf_process_exec_event *)ev;
+		if (exec->pids.tid != exec->pids.tgid)
```
nicholasberlin, line 167: Why this change?

nicholasberlin: So exec'ing, changes the tid to be the tgid?

haesbaert: basically exec always comes from tgid, so tid is always equal tgid.
even if you do an exec from a child thread, at some point control passes to the main thread and there is where the exec happens.

haesbaert: I think it's basically here:

https://elixir.bootlin.com/linux/v6.12.6/source/fs/exec.c#L1063-L1068

nicholasberlin (approved):

bpf_queue.c
nicholasberlin: Thanks, I was trying to find that.

PR #118, opened 2025-02-04
Make sure we always get tgid from exit

haesbaert:
Nasty bug introduced by me trying to be overzealous.

When an exit happens, it's undefined if it will come from the group leader or not, it's not like exec.

We were dropping exits that didn't come from the leader, which is very evil. The probes are careful enough to only send one exit event once the group leader died, but that doesn't mean it comes _FROM_ the group leader, so we have to look at tgid.

Testing this is a bit tricky, since it's pretty much random, we brute force it, which makes the test a bit slow, but should be improved later.

We can't use pthread because we build statically, so we have to go directly through clone(2). I lost a lot of time thinking I had to to use clone3(2) directly, only to realize I could use clone(2) where libc already tidies up the stack.

### DIFF
--- a/quark-test.c
+++ b/quark-test.c
@@ -258,7 +258,7 @@ fork_clone_and_exit(void)
 	/* continue on child ... */
 
 	/*
-	 * Set up a stack, clone3() is like fork and we just give it a stack
+	 * Set up a stack, clone() is like fork and we just give it a stack
 	 * without a starting addr.
 	 */
 	stack = mmap(NULL, STACK_SIZE, PROT_READ | PROT_WRITE,
--- a/quark-test.c
+++ b/quark-test.c
@@ -6,6 +6,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <sched.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


quark-test.c
```diff
@@ -225,6 +228,73 @@ fork_exec_nop(void)
...
+	    CLONE_SYSVSEM;
+
+	/*
+	 * clone the new thread, pid is tid
+	 */
+	pid = clone(clone_start, stack_start, flags, NULL);
+	if (pid == -1)
+		err(1, "clone3");
+
+	/* Wait for the child to exit */
```
nicholasberlin, line 285:
```suggestion
	/* Wait for the clone thread to exit */
```

quark-test.c
```diff
@@ -225,6 +228,73 @@ fork_exec_nop(void)
...
+	 */
+	if ((pid = fork()) == -1)
+		err(1, "fork");
+	/* parent just returns */
+	if (pid != 0)
+		return (pid);
+	/* continue on child ... */
+
+	/*
+	 * Set up a stack, clone3() is like fork and we just give it a stack
```
nicholasberlin, line 261:
```suggestion
	 * Set up a stack, clone() is like fork and we just give it a stack
```

nicholasberlin (approved):

Issue #113, opened 2025-01-22
Kill initial snapshot


haesbaert:
The initial snapshot doesn't make more sense how that there are proper process iterators.
With the addition of sockets, the snapshot becomes weirder, we would have to send all processes then all sockets.

It makes more sense to let the user get the snapshot when/if he wants via `quark_process_iter_next` and `quark_socket_iter_next`, it's virtually no additional work for the user and it cuts a ton of code.

PR #122, opened 2025-02-11
Zap initial snapshots. Issue #113

haesbaert:
Zap initial snapshots. Issue #113
This removes the initial snapshot of processes that it was sent on queue bootup.
The user is now expected to use quark_queue_iter(3) to retrieve them.

This simplified the code quite a bit, and as we add more kind of events,
deciding what to send initially becomes silly. We provide decent iterators and
users rejoice.

Documentation follows on the next commit.

### DIFF
--- a/quark_process_iter.3
+++ b/quark_process_iter.3
@@ -1,0 +1,42 @@
+.Dd $Mdocdate$
+.Dt QUARK_PROCESS_ITER 3
+.Os
+.Sh NAME
+.Nm quark_process_iter_init ,
+.Nm quark_process_iter_next
+.Nd string process iteration operations
+.Sh SYNOPSIS
+.In quark.h
+.Ft void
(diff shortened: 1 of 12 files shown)


quark_process_iter.3
```diff
@@ -0,0 +1,42 @@
...
+.Nd string process iteration operations
+.Sh SYNOPSIS
+.In quark.h
+.Ft void
+.Fn quark_process_iter_init "struct quark_process_iter *" "struct quark_queue *qq"
+.Ft const quark_process *
+.Fn quark_process_iter_next "struct quark_process_iter *"
+.Sh DESCRIPTION
+Initialize and iterate over the internal cache of processes, it can be used on
+boot up to get a list of all learned processes, or else.
```
nicholasberlin, line 16: or else what?

nicholasberlin (approved):

PR #131, opened 2025-03-13
Add process iterator test

Tacklebox:


quark-test.c
```diff
@@ -436,6 +436,103 @@ t_fork_exec_exit(const struct test *t, struct quark_queue_attr *qa)
...
+
+	if (quark_queue_open(&qq, qa) != 0)
+		err(1, "quark_queue_open");
+
+	quark_process_iter_init(&qi, &qq);
+
+	for (i = 0; i < n_children; i++) {
+		children[i] = fork_exec_nop();
+		drain_for_pid(&qq, children[i]);
+	}
```
haesbaert, line 461:
```suggestion
	for (i = 0; i < (int)nitems(children); i++)
		children[i] = fork_exec_nop();
	for (i = 0; i < (int)nitems(children); i++)
		children[i] = drain_for_pid(&qq, children[i]);
```

haesbaert: Small trick, use nitems() so you can cut n_children out.
Because of how aging works in quark, it makes more sense to fork all + drain all, otherwise each drain will have to age the event for 25ms, so this cuts 75ms->25ms

quark-test.c
```diff
@@ -436,6 +436,103 @@ t_fork_exec_exit(const struct test *t, struct quark_queue_attr *qa)
...
 }
 
+static int
+t_process_iterator(const struct test *t, struct quark_queue_attr *qa)
+{
+	struct quark_queue		 qq;
+	struct quark_process_iter	 qi;
+	const struct quark_process	*qp;
+	int				 n_children = 3;
+	pid_t				 children[n_children];
```
haesbaert, line 446:
```suggestion
	pid_t				 children[3];
```
See the nitems() comment below, also .

quark-test.c
```diff
@@ -436,6 +436,103 @@ t_fork_exec_exit(const struct test *t, struct quark_queue_attr *qa)
...
+	size_t				 expected_len;
+	int				 i;
+	int				 j;
+	char				 cwd[PATH_MAX];
+
+	if (quark_queue_open(&qq, qa) != 0)
+		err(1, "quark_queue_open");
+
+	quark_process_iter_init(&qi, &qq);
+
```
haesbaert, line 457: Here you initialize the iterator before drain_for_pid, so the iterator can be invalidated, just move this to before the  main loop. I realize this is not ideal, maybe I should change iter_init to point to NULL, instead of pointing to RB_ROOT.

quark-test.c
```diff
@@ -436,6 +436,103 @@ t_fork_exec_exit(const struct test *t, struct quark_queue_attr *qa)
...
+		*/
+		for (expected_len = args->argc, j = 0; j < args->argc; j++)
+			expected_len += strlen(args->argv[j]);
+		args_free(args);
+		assert(qp->cmdline_len == expected_len);
+
+		if (getcwd(cwd, sizeof(cwd)) == NULL)
+			err(1, "getcwd");
+		assert(!strcmp(cwd, qp->cwd));
+
```
haesbaert, line 529:
```suggestion

```

I think we can cut this, this is already tested multiple times in other places, if we find the processes through iteration I'm happy.

haesbaert (changes requested):

quark-test.c
```diff
@@ -436,6 +436,103 @@ t_fork_exec_exit(const struct test *t, struct quark_queue_attr *qa)
...
+	for (i = 0; i < n_children; i++) {
+		children[i] = fork_exec_nop();
+		drain_for_pid(&qq, children[i]);
+	}
+
+	for (i = 0; i < n_children; i++) {
+		do {
+			qp = quark_process_iter_next(&qi);
+		} while (qp != NULL && (pid_t)qp->pid != children[i]);
```
haesbaert, line 466: The outer loop should be the iterator, and then look for the children.
This is only working now because of how the internal tree is sorted, if that changes, you would be skipping elements.

PR #149, opened 2025-04-07
Fedora tests (aka make the ci great again^W)

haesbaert:
This improves the CI by running quark-test on fedora versions 28-41.

krun-fedora.sh downloads the latest kernel rpm of the specified version, unpacks
and runs krun.sh on it. When Fedora updates its rpms, we end up getting the
updated version.

This also enables nested virtualization on the CI so we can run things on qemu
super fast, <1min as opposed to 5-10min in the worst case, the most important
bit is that it fixes some flaky tests that would time out.

Prompted by a question from @andrewkroh

Many thanks to @mjwolf for pointing me to the allowed GCP options in buildkite.

### DIFF
--- a/krun-fedora.sh
+++ b/krun-fedora.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+set -euo pipefail
+
+SCRIPT=${0##*/}
+
+function usage
+{
+	echo "usage: $SCRIPT initramfs.gz FEDORAVERSION command" 1>&2
+	exit 1
+}
+
+if [ $# -lt 3 ]; then
+	usage
+fi
+
+INITRAMFS="$1"
+FEDORAVER="$2"
+shift 2
+
+case $FEDORAVER in
+2?|3?)	URL="https://archives.fedoraproject.org/pub/archive/fedora/linux/updates/$FEDORAVER/Everything/x86_64/Packages/k";;
+4?)	URL="http://ftp.fau.de/fedora/linux/updates/$FEDORAVER/Everything/x86_64/Packages/k";;
+*)	echo bad version "$FEDORAVER" 1>&2
+esac
+
+TMPDIR=$(mktemp -d "/tmp/$SCRIPT.XXXXXXXXXX")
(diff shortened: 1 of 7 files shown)


krun-fedora.sh
```diff
@@ -0,0 +1,47 @@
...
+
+set -euo pipefail
+
+SCRIPT=${0##*/}
+TMPDIR=$(mktemp -d "/tmp/$SCRIPT.XXXXXXXXXX")
+
+function usage
+{
+	echo "usage: $SCRIPT initramfs.gz FEDORAVERSION command" 1>&2
+	exit 1
```
nicholasberlin, line 11: I think `TMPDIR` won't be cleaned up if we exit here. Perhaps use trap:
```
trap 'rm -rf "$TMPDIR"' EXIT
```

nicholasberlin: Here's a reusable bash script header from the AIs:
```
#!/bin/bash
set -euo pipefail

SCRIPT_NAME="${0##*/}"

# Create and track a temp directory
TMPDIR=$(mktemp -d "/tmp/$SCRIPT_NAME.XXXXXXXXXX")

# Cleanup function
cleanup() {
    local exit_code=$?
    echo "[$SCRIPT_NAME] Cleaning up (exit code: $exit_code)..."
    rm -rf "$TMPDIR"
}

# Error handler
on_error() {
    echo "[$SCRIPT_NAME] Error on line $LINENO: $BASH_COMMAND"
}

# Traps
trap cleanup EXIT
trap on_error ERR

echo "[$SCRIPT_NAME] Using temp dir: $TMPDIR"

# --- your script logic here ---


```

haesbaert: > I think `TMPDIR` won't be cleaned up if we exit here. Perhaps use trap:
> 
> ```
> trap 'rm -rf "$TMPDIR"' EXIT
> ```

well spotted, zefixed, but without AI bloat. Many thanks!

nicholasberlin (approved):

PR #151, opened 2025-04-08
RHEL CI Fiesta

haesbaert:
Add 15 RHEL releases to CI

Same dance as fedora, add a new script krun-rhel.sh.
Since it's all pretty much the same, I'll consider merging it into
krun-distro.sh in the future, especially when we add suse.

It all depends how adding debian and ubuntu look, so keep it separate from now.

While here, use https on fedora mirrors.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,7 +40,7 @@ steps:
 
   - label: "quark-test on fedora 28 (no bpf)"
     key: test_fedora_28
-    command: "./.buildkite/runtest_fedora.sh 28 -k"
+    command: "./.buildkite/runtest_distro.sh fedora 28 -k"
     depends_on:
       - make_docker
     agents:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -194,7 +194,172 @@ steps:
...
+    command: "./.buildkite/runtest_distro.sh rhel 8.9"
+    depends_on:
+      - make_docker
+    agents:
+      image: family/core-ubuntu-2204
+      provider: gcp
+      machineType: n2-standard-2
+      enableNestedVirtualization: true
+
+  - label: "quark-test on rhel 9"
```
nicholasberlin, line 294: what the difference between this and rhel 9.0? perhaps the script will show me.

nicholasberlin (approved):

.buildkite/pipeline.yml
haesbaert: 9 is the current 9 mainline, 9.0 is the original 9.0 release.
5.14.0-503.35.1.el9_5.x86_64
vs
4.18.0-425.3.1.el8.x86_64

haesbaert: ooops, just realized the URL for 8.x and 9.x was incorrect :(
812325c

PR #156, opened 2025-04-25
Get rid of cdefs.h

haesbaert:
cdefs.h is old school and marked as deprecated in musl. It seems nothing really needed, so just zap it.


haesbaert:
> This change is to allow native builds on non-glibc systems, right? Would the `centos7` build still work?

yes and yes :)
we do test centos7 build in the ci: https://buildkite.com/elastic/quark/builds/280#01966c93-217f-46fc-b80c-514b44f9025b

PR #171, opened 2025-05-14
Implement entity_id. Issue #155

haesbaert:
~~This implements the 12 byte entity id as defined in https://www.elastic.co/docs/reference/ecs/ecs-process#field-process-entity-id~~

~~We can't depend on dynamic linking of md or openssl, so include a standlone MIT licensed sha256 implementation from https://github.com/ilvn/SHA256, they claim to be formally verified, so that's something.~~

~~We now also have to link against resolv so we can get the base64 functions, that's ok, beats already links against it.~~

~~This is a **WIP** as I want to make sure we compute the very same entity_id as gosysinfo and friends.~~

```
    Implement entity_id. Issue #155

    This implements entity id as defined in
    https://www.elastic.co/docs/reference/ecs/ecs-process#field-process-entity-id

    I don't want to link with libresolv just to get b64_ntop, so include our own
    portable version (musl also doesn't have it).

    Entity id's only requirement is to be unique, so use pid number + boot time,
    which should be enough.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,7 @@ LIBQUARK_DEPS+= $(EEBPF_FILES) include
 endif
 LIBQUARK_DEPS:= $(filter-out manpages.h, $(LIBQUARK_DEPS))
 LIBQUARK_SRCS:=			\
+	base64.c		\
 	bpf_queue.c		\
 	btfhub.c		\
 	compat.c		\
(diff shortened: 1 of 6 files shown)


haesbaert:
I think I can use a weak symbol from md or openssl, it seems they both provide strong symbols, so by including ours as a weak symbol it should work as a fallback.
Also, no b64_ntop in musl, so have to provide one.

haesbaert:
I've decided to go for a stupidly simple entity_id, the only requirement is too be unique, so pid + boot time should be enough in linux, no point in trying to match other implementations as everyone does differently anyway.

PR #175, opened 2025-05-29
Make a few compat symbols weak

nicholasberlin:


nicholasberlin:
Thanks for the updates, my local compile worked.

PR #183, opened 2025-06-04
Make quark-test valgrind friendly

haesbaert:
This took me more time to figure out than I care to admit.

The tests that did fork+exec would blow up under valgrind with weird outout, kprobe would even segfault.

The fork+exec was done with the binary retrieved via binpath() which reads /proc/self/exe.

Valgrind is not a monitor that forks+execs and ptraces and whatnot, it's an actual VM, so /proc/self/exe was pointing back to a valgrind tool! Ironically, we would run quark-test under valgrind which in turn would try to run valgrind in itself.

Fix it by changing how we compute our binary path, use argv[0] + realpath(3). If you're wondering why we do the dance of forking and execing self, it's because when we run from an initramfs, we have no other binaries to run, so quark-test is "fully" self contained.

Part of an effort to add valgrind runs to the CI.

### DIFF
--- a/quark-test.c
+++ b/quark-test.c
@@ -108,7 +108,7 @@ binpath(const char *bin)
 	static char	name[PATH_MAX];
 
 	if (bin != NULL && realpath(bin, name) == NULL)
-		err(1, "can' initialize binpath");
+		err(1, "can't initialize binpath");
 	else if (bin == NULL && name[0] == 0)
 		err(1, "binpath not initialized");
 
--- a/quark-test.c
+++ b/quark-test.c
@@ -103,14 +103,16 @@ color(int color)
 }
 
 static char *
-binpath(void)
+binpath(const char *bin)
 {
 	static char	name[PATH_MAX];
 
-	if (readlink("/proc/self/exe", name, sizeof(name)) == -1)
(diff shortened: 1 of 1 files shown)


quark-test.c
```diff
@@ -103,14 +103,16 @@ color(int color)
...
 static char *
-binpath(void)
+binpath(const char *bin)
 {
 	static char	name[PATH_MAX];
 
-	if (readlink("/proc/self/exe", name, sizeof(name)) == -1)
-		err(1, "readlink");
+	if (bin != NULL && realpath(bin, name) == NULL)
+		err(1, "can' initialize binpath");
```
nicholasberlin, line 111:
```suggestion
		err(1, "can't initialize binpath");
```

quark-test.c
```diff
@@ -103,14 +103,16 @@ color(int color)
...
+	else if (bin == NULL && name[0] == 0)
```
nicholasberlin, line 112: if bin is null, `name[0]` will have stack garbage in it.

nicholasberlin: Oh, nevermind, it's static.

nicholasberlin (approved):

PR #186, opened 2025-06-05
Load tty_write probe for BYPASS

nicholasberlin:


haesbaert:
I'd rather have QQ_TTY that only works on bypass, I say this because in the future I think tty events are also interesting outside of bypass. Currently setsid is there because my plan is to make it default outside of bypass as well (basically all process things are default).
What do you think? I'll push a change, here, and if you don't like it I can remove it.

haesbaert:
Pushed, also added a tiny test that at least tells is if the tty probe loads

nicholasberlin:
Make sense to me, I realized we don't always want to load those probes for BYPASS.

Thanks! LGTM

haesbaert (approved):

Issue #150, opened 2025-04-08
Write krun-rhel.sh and krun-ubuntu.sh


haesbaert:
See krun-fedora.sh, we need similar things.

krun-rhel in #151

stanek-michal:
https://github.com/elastic/quark/pull/192

PR #192, opened 2025-06-13
krun-ubuntu initial version

stanek-michal:
needs review, arm64 not tested yet

issue: https://github.com/elastic/quark/issues/150

### DIFF
--- a/krun-ubuntu.sh
+++ b/krun-ubuntu.sh
@@ -1,0 +1,169 @@
+#!/usr/bin/env bash
+#
+# krun-ubuntu.sh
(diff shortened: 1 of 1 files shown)


haesbaert:
Some more points
 - The temporary files are not deleted, they should
 - Can you hide the [INFO] prints under a `-v`, we can add the same later on the other ones.
 - Indentation should be tabs, not two spaces

stanek-michal:
Looks great, I squashed the commits. You can ACK and merge

haesbaert (approved):

PR #201, opened 2025-07-02
Only attempt to probe ipv6 if present

nicholasberlin:


quark.c
```diff
@@ -2065,9 +2065,12 @@ sproc_scrape(struct quark_queue *qq)
 		r = sproc_net_tcp(qq, AF_INET, &socket_tmp_tree);
 		if (r == -1)
 			goto done;
-		r = sproc_net_tcp(qq, AF_INET6, &socket_tmp_tree);
-		if (r == -1)
-			goto done;
+		if (ipv6_supported())
+		{
```
haesbaert, line 2069:
```suggestion
		if (ipv6_supported()) {
```

qutil.c
```diff
@@ -215,6 +215,21 @@ load_file_nostat(int fd, size_t *total)
...
+
+	fp = fopen("/proc/sys/net/ipv6", "r");
+	if (fp != NULL) {
+		enabled = 1;
+		fclose(fp);
+	}
+
+	return enabled;
+}
+
```
haesbaert, line 232: I'm ok with this but a part of me thinks we should be trying to create an AF_INET6 socket instead of opening /proc.
like
```
fd = socket(AF_INET6, SOCK_STREAM, 0)
enable = fd != -1
```

haesbaert (approved):
the syntax on if (ipv6_supported()... needs fixing, otherwise fine.

nicholasberlin:
Thanks for the review. I've tested this locally with a kernel built to load ipv6 as a module and the tests pass. Added an issue to figure out to add a kernel like this to CI: https://github.com/elastic/quark/issues/202

haesbaert:
Can I squash and merge this?

PR #207, opened 2025-07-27
Change quark-test to use /bin/true instead of itself

haesbaert:
Basically our tests fork and exec itself with a -N (nop) option, we then look at process events and check if it's what we wanted.

This all worked great before as we used to figure out who we were by readlink(/proc/self/exe), but then I introduced the valgrind tests, where /proc/self/exe is some valgrind binary, not "us" (documented in bbf265bb864edece8e857db9f22cd7df57479a0b), so I changed the way we figure "us" by doing realpath(argv0), which is also bad.

The problem with realpath(argv0) is that it's not sufficient if quark-test was executed via some PATH directory, just copy quark-test into $HOME/bin, and then cd /tmp && quark-test and it will fail, as binpath() won't find it.

So stop this madness, use /bin/true for nop, instead of execing ourselves, which involves us also adding a /bin/true to the initramfs, so one is provided.

### DIFF
--- a/quark-test.c
+++ b/quark-test.c
@@ -105,19 +105,6 @@ color(int color)
 	return (ret);
 }
 
-static char *
-binpath(const char *bin)
-{
-	static char	name[PATH_MAX];
-
-	if (bin != NULL && realpath(bin, name) == NULL)
-		err(1, "can't initialize binpath");
-	else if (bin == NULL && name[0] == 0)
-		err(1, "binpath not initialized");
-
-	return (name);
-}
-
 static int
 backend_of_attr(struct quark_queue_attr *qa)
 {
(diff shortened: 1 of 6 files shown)


quark-test.c
```diff
@@ -665,8 +649,9 @@ t_fork_exec_exit(const struct test *t, struct quark_queue_attr *qa)
 	assert(qp->proc_tty_minor != 0);
 #endif
 	/* check strings */
-	assert(!strcmp(qp->comm, program_invocation_short_name));
-	assert(!strcmp(qp->filename, binpath(NULL)));
+	assert(!strcmp(qp->comm, "true"));
+	assert(!strcmp(qp->filename, "/bin/true") ||
+	    !strcmp(qp->filename, "/usr/bin/true"));
```
nicholasberlin, line 654: Is this side of the `||` needed?

nicholasberlin (approved):

quark-test.c
haesbaert: I mean, when you run in the host, outside of initramfs, it can be /usr/bin/true

PR #210, opened 2025-08-01
Resolve relative exec filename in bpf probes

nicholasberlin:
The filename resolution within elastic/ebpf's exec probe can return relative paths:
```
->2804617 (FORK+EXEC+EXIT)
  COMM  comm=whoami
  CMDL  cmdline=[ ../usr/bin/whoami ]
  CWD   cwd=/tmp
  FNAM  filename=../usr/bin/whoami

->2804726 (FORK+EXEC+EXIT)
  COMM  comm=whoami
  CMDL  cmdline=[ ../usr/share/../bin/whoami ]
  CWD   cwd=/tmp
  FNAM  filename=../usr/share/../bin/whoami
```

This PR updates the probes to resolve the path using the internal function `ebpf_resolve_path_to_string` instead of copying the contents of `binprm->filename`.

Additionally, an ebpf-only test (kprobes still suffers this problem) was added.
To facilitate testing, the Makefile was modified for the `true` binary to be used for local and qemu testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -347,7 +347,7 @@ test-valgrind: quark-test
 initramfs:
 	mkdir -p initramfs/bin
 
-initramfs.gz: init quark-mon-static quark-btf-static quark-test-static true initramfs
+initramfs.gz: init quark-mon-static quark-btf-static quark-test-static initramfs
 	$(call assert_no_syslib)
 	cp init initramfs/
 	cp quark-mon-static initramfs/quark-mon
@@ -375,7 +375,7 @@ quark-btf: quark-btf.c manpages.h $(LIBQUARK_TARGET)
 	$(Q)$(CC) $(CFLAGS) $(CPPFLAGS) $(CDIAGFLAGS) \
 		-o $@ $< $(LIBQUARK_TARGET) $(EXTRA_LDFLAGS)
 
-quark-test: quark-test.c manpages.h $(LIBQUARK_TARGET)
+quark-test: quark-test.c manpages.h $(LIBQUARK_TARGET) true
 	$(call msg,CC,$@)
 	$(Q)$(CC) $(CFLAGS) $(CPPFLAGS) $(CDIAGFLAGS) \
 		-o $@ $< $(LIBQUARK_TARGET) $(EXTRA_LDFLAGS)
@@ -392,7 +392,7 @@ quark-btf-static: quark-btf.c manpages.h $(LIBQUARK_STATIC_BIG)
 	$(Q)$(CC) $(CFLAGS) $(CPPFLAGS) $(CDIAGFLAGS) \
 		-static -o $@ $< $(LIBQUARK_STATIC_BIG) $(EXTRA_LDFLAGS)
 
-quark-test-static: quark-test.c manpages.h $(LIBQUARK_STATIC_BIG)
+quark-test-static: quark-test.c manpages.h $(LIBQUARK_STATIC_BIG) true
 	$(call msg,CC,$@)
 	$(call assert_no_syslib)
 	$(Q)$(CC) $(CFLAGS) $(CPPFLAGS) $(CDIAGFLAGS) \
@@ -474,6 +474,7 @@ clean:
 		quark-btf-static	\
 		quark-test		\
 		quark-test-static	\
+		true			\
 		btf_prog_skel.h		\
 		init
 	$(Q)rm -rf initramfs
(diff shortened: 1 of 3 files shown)


.buildkite/runtest.sh
```diff
@@ -15,5 +15,8 @@ fi
 download quark-test .
 chmod +x quark-test
 
+download true .
+chmod +x true
+
```
haesbaert, line 20:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -14,6 +14,7 @@ steps:
     command: "make centos7"
     artifact_paths:
       - "quark-test"
+      - "true"
```
haesbaert, line 17:
```suggestion
```

Dockerfile.centos7
```diff
@@ -2,4 +2,4 @@ FROM centos:7
 RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/CentOS-*.repo
 RUN sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/CentOS-*.repo
 RUN sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/CentOS-*.repo
-RUN yum install -y gcc make m4 less
+RUN yum install -y gcc make m4 less glibc-static
```
haesbaert, line 5:
```suggestion
RUN yum install -y gcc make m4 less
```

Makefile
```diff
@@ -347,7 +347,7 @@ test-valgrind: quark-test
 initramfs:
 	mkdir -p initramfs/bin
 
-initramfs.gz: init quark-mon-static quark-btf-static quark-test-static true initramfs
+initramfs.gz: init quark-mon-static quark-btf-static quark-test-static initramfs
```
haesbaert, line 350:
```suggestion
initramfs.gz: init quark-mon-static quark-btf-static quark-test-static true initramfs
```

Makefile
```diff
@@ -375,7 +375,7 @@ quark-btf: quark-btf.c manpages.h $(LIBQUARK_TARGET)
 	$(Q)$(CC) $(CFLAGS) $(CPPFLAGS) $(CDIAGFLAGS) \
 		-o $@ $< $(LIBQUARK_TARGET) $(EXTRA_LDFLAGS)
 
-quark-test: quark-test.c manpages.h $(LIBQUARK_TARGET)
+quark-test: quark-test.c manpages.h $(LIBQUARK_TARGET) true
```
haesbaert, line 378:
```suggestion
quark-test: quark-test.c manpages.h $(LIBQUARK_TARGET)
```

Makefile
```diff
@@ -392,7 +392,7 @@ quark-btf-static: quark-btf.c manpages.h $(LIBQUARK_STATIC_BIG)
 	$(Q)$(CC) $(CFLAGS) $(CPPFLAGS) $(CDIAGFLAGS) \
 		-static -o $@ $< $(LIBQUARK_STATIC_BIG) $(EXTRA_LDFLAGS)
 
-quark-test-static: quark-test.c manpages.h $(LIBQUARK_STATIC_BIG)
+quark-test-static: quark-test.c manpages.h $(LIBQUARK_STATIC_BIG) true
```
haesbaert, line 395:
```suggestion
quark-test-static: quark-test.c manpages.h $(LIBQUARK_STATIC_BIG)
```

quark-test.c
```diff
@@ -330,7 +330,12 @@ fork_exec_nop(void)
 		err(1, "fork");
 	else if (child == 0) {
 		/* child */
-		return (execvp("true", argv));
+		if (relative) {
+			if (execvp("bin/true", argv) == -1)
+				return execvp("./true", argv);
+		}
+		else
+			return execvp("true", argv);
```
haesbaert, line 338: While testing relative is good, this is a bit too complex, there's no need to copy true around, we can do the same with chdir, I'll push a change

haesbaert:
Heya I did some changes, I like it but I think we can do without relying on having a local true always built, this would also break when/if you copy quark-test to /usr/bin and run quark-test from there. So I removed them.

Basically the same as you were doing but instead just chdir into /bin or /usr/bin and run it relative there.
Also normalized the naming a bit.

I'm not sure why you added glibc-static to the centos7 packages, as we don't build initramfs.gz,

PR #213, opened 2025-08-08
Add cJSON ifndef SYSLIB

haesbaert:
Preparation for the upcoming kubernetes support.
cJSON has some limitations but it's straightforward and easy to ship.


Tacklebox (approved):
LGTM, just double check that the cJSON license/authors are not supposed to go into the notice.txt

haesbaert:
> LGTM, just double check that the cJSON license/authors are not supposed to go into the notice.txt

So in the past I had everything in notice, but then was instructed it wasn't necessary anymore:
https://github.com/elastic/open-source/issues/435#issuecomment-2250598390

I've added the MIT license to the licenses/ directory.

PR #214, opened 2025-08-08
Add an API for starting upcoming quark-kube-talker

haesbaert:
This is more preparatory work for kubernetes support.

We have to be flexible, this means being able to run on an environment without multithreading, in an environment with multithreading, and in some gc runtime like golang.

The object is to have "something" that feeds a file descriptor with kubernetes events, this "something" is the upcoming quark-kube-talker.

We tell quark that kubernetes is active by setting up a quark-kube-talker (or whatever), and feeing quark_queue_open with the pipe to read from. For now, management of the filedescriptor is done by the caller, so it must start quark-kube-talker, get the fd, and close on its own.

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -2431,6 +2431,7 @@ quark_queue_default_attr(struct quark_queue_attr *qa)
 	qa->max_length = 10000;
 	qa->cache_grace_time = 4000;	/* four seconds */
 	qa->hold_time = 1000;		/* one second */
+	qa->kubefd = -1;		/* disabled */
 }
 
 int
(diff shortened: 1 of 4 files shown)


quark.c
```diff
@@ -2897,3 +2898,42 @@ quark_queue_get_event(struct quark_queue *qq)
...
+
+        /* child */
+        close(pipefd[0]);
+
+        if (qclosefrom(3, pipefd[1]) == -1) {
+                qwarn("qclosefrom");
+                qwarnx("not closing child descriptors");
+        }
+
+        snprintf(pidbuf, sizeof(pidbuf), "%d", pipefd[1]);
```
nicholasberlin, line 2932: why named `pidbuf`? it's filled with a file descriptor, right?

haesbaert: whoooooooooooooops, it used to be a pid, and the pipe was on another buffer, I changed and renamed the wrong one, I'll fix it, I also realized, that I had my tabs disabled in emacs and the whole function is not using tabs. :D

PR #227, opened 2025-09-23
Instead of hardcoding /sys/fs/cgroup, search.

nicholasberlin:
There are occasions where older systems which support ebpf have both cgroup v1 and cgroup v2 mounted. This commit adds new functionality to parse /proc/self/mountinto and iterate all the mount points. Each mount is opened and checked for cgroup2, and if confirmed the iteration completes. Iteration is preferable to parsing the filesystem field because paths can be overmounted.

Update init.c to mount cgroup2 in a non-standard location with spaces in the directory name for testing.

### DIFF
--- a/bpf_queue.c
+++ b/bpf_queue.c
@@ -4,6 +4,7 @@
 #include <sys/epoll.h>
 #include <sys/param.h>
 #include <sys/sysinfo.h>
+#include <sys/vfs.h>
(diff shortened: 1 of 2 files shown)


nicholasberlin:
Closing in favor of: https://github.com/elastic/quark/pull/228

PR #228, opened 2025-09-24
Try harder to get a cgroup2 root mounting point

haesbaert:
his looks up the path for cgroup2, in case it's not in /sys/fs/cgroup, but more
importantly, it makes sure /sys/fs/cgroup is a cgroup2, and if not, it attempts
to mount a temporary cgroup2 for loading the DNS probes.

On some systems, cgroup2 is not mounted by default, even though the kernel
supports it.

For DNS and/or every future cgroup probe to function, we need to fetch the
cgroup mounting point pass it to bpf_probes__load(). After that the fd is not
really needed, we can close and also unmount the temporary cgroup2 mount we
might have created.

Idea from Nick.

### DIFF
--- a/bpf_queue.c
+++ b/bpf_queue.c
@@ -3,6 +3,7 @@
 
 #include <sys/epoll.h>
 #include <sys/param.h>
+#include <sys/mount.h>
 #include <sys/sysinfo.h>
 
 #include <ctype.h>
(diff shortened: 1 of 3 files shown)


bpf_queue.c
```diff
@@ -590,6 +694,7 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
 	struct bpf_probes		*p;
 	struct ring_buffer_opts		 ringbuf_opts;
 	int				 cgroup_fd, i, off, ringbuf_fd;
+	char				*cgroup_unmount;
```
nicholasberlin, line 697: weird spacing here

bpf_queue.c
```diff
@@ -526,6 +527,109 @@ bpf_ringbuf_cb(void *vqq, void *vdata, size_t len)
...
+		if ((end = strchr(start, ' ')) == NULL) {
+			qwarn("no space 2");
+			continue;
+		}
+		path = strndup(start, end - start);
+		break;
+	}
+
+	/*
+	 * No cgroup2 mount found, try mountint it ourselves at
```
nicholasberlin, line 569:
```suggestion
	 * No cgroup2 mount found, try to mount it ourselves at
```

bpf_queue.c
```diff
@@ -526,6 +527,109 @@ bpf_ringbuf_cb(void *vqq, void *vdata, size_t len)
...
+	 * /tmp/quark_cgroup2_mount.XXXXX.
+	 * If we succeed, we must pass the mounting point up to the caller, so
+	 * it can umount after the probes are loaded.
+	 */
+	if (path == NULL) {
+		char template[] = "/tmp/quark_cgroup2_mount.XXXXXX";
+
+		if ((path = mkdtemp(template)) == NULL) {
+			qwarn("mktemp %s", template);
```
nicholasberlin, line 578:
```suggestion
			qwarn("mkdtemp %s", template);
```

bpf_queue.c
```diff
@@ -526,6 +527,109 @@ bpf_ringbuf_cb(void *vqq, void *vdata, size_t len)
...
+		if (path == NULL) {
+			qwarn("strdup");
+			goto fail;
+		}
+		if (mount(NULL, path, "cgroup2", 0, NULL) == -1) {
+			qwarn("mount %s", path);
+			goto fail;
+		}
+
+		*unmount_path = strdup(path);
```
nicholasberlin, line 594: we check for `strdup` failures on line 585. should make them consistent. if it fails (and all other code miraculously works) we will fail to umount the custom cgroup2 mount.

bpf_queue.c
```diff
@@ -526,6 +527,109 @@ bpf_ringbuf_cb(void *vqq, void *vdata, size_t len)
...
+	}
+
+	if (path == NULL) {
+		qwarnx("no cgroup2 mount");
+		goto fail;
+	}
+
+	if ((fd = open(path, O_RDONLY)) == -1) {
+		qwarn("open %s", path);
+		goto fail;
```
nicholasberlin, line 604: `*unmount_path` could be set here, which is fine since we `cgroup2_umount_tmp` in the calling function's `fail:` (line 965).

could set `*unmount_path` to NULL at the beginning of this function, and conditionally `umount` and `free` in this function's `fail:` code. in that way failures in this function are cleaned up, and line 965 is for cleaning up that function.

/shrug

bpf_queue.c
haesbaert: I thought about it, it would basically just leave it dangling, while in practice that strdup will never fail, so I just ignored. I can add some print at least

bpf_queue.c
haesbaert: zefixed

bpf_queue.c
haesbaert: ditto

bpf_queue.c
```diff
@@ -526,6 +527,114 @@ bpf_ringbuf_cb(void *vqq, void *vdata, size_t len)
...
+	}
+	free(path);
+
+	return (fd);
+
+fail:
+	free(path);
+	if (fd != -1)
+		close(fd);
+	cgroup2_umount_tmp(umount_path);
```
nicholasberlin, line 633: Nice.

nicholasberlin (approved):
LGTM

Issue #233, opened 2025-10-07
Compute and add session view fields in ecs


haesbaert:
We have the source data, just need to compute and add to the output.

PR #235, opened 2025-10-07
Add session view data to ECS. Issue #233

haesbaert:
This adds the following fields to ECS:
 - process.group_leader.*
 - process.session_leader.*
 - process.entry_leader.*
 - process.entry_leader.entry_meta.type

group_leader and session_leader are the quark_process of proc_pgid and proc_sid, respectively. There's a boolean that is true if they are the same as the top level process, which is the actual process we are reporting (process.*).

The hard part was already done last year, which is to compute entry_leader, the computation has to be opted in via QQ_ENTRY_LEADER.

There's an proecss.entry_leader.entry_meta.type which I didn't find the ECS docs, but it's required by kibana, so I copied what the session view processor in beats is doing.

This hasn't been tested in Kibana, so I might have missed something, but the output looks good if I compare against cloud defend.


ecs.c
```diff
@@ -592,6 +607,84 @@ ecs_process(struct quark_queue *qq, struct hanson *h,
...
+		hanson_add_object(h, "entry_leader", first);
+		{
+			int	entry_leader_first = 1;
+
+			ecs_process(qq, h, entry_leader,
+			    ECS_PROC_LEADER, &entry_leader_first);
+			hanson_add_key_value_bool(h, "same_as_process",
+			    entry_leader->pid == qp->pid, &entry_leader_first);
+			/* process.entry_leader.entry_meta.* */
+			hanson_add_object(h, "entry_meta", first);
```
nicholasberlin, line 675: should `first` be `entry_leader_first`?

PR #250, opened 2025-10-28
Add bytes_received and bytes_sent to quark_socket

nicholasberlin:


quark.h
```diff
@@ -275,6 +275,8 @@ struct raw_sock_conn {
 	struct quark_sockaddr	local;
 	struct quark_sockaddr	remote;
 	enum sock_conn		conn;
+	uint64_t	bytes_received;
+	uint64_t	bytes_sent;
```
haesbaert:
```suggestion
	u64	bytes_received;
	u64	bytes_sent;
```

quark.c
```diff
@@ -4432,14 +4436,19 @@ raw_event_sock(struct quark_queue *qq, struct raw_event *raw)
...
+			    raw->pid, raw->time);
 			if (qsk == NULL) {
 				qwarn("socket_alloc");
 				return (NULL);
 			}
 		}
 		if (qsk->close_time == 0)
 			qsk->close_time = raw->time;
+		if (qsk->bytes_received == 0 && qsk->bytes_sent == 0) {
+			qsk->bytes_received = conn->bytes_received;
```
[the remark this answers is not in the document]
nicholasberlin: I was copying the block above without really thinking ...

Setting it always makes sense. It will overwrite the values set in the alloc function, but that's fine.

quark.h
nicholasberlin: forced pushed the change: [4187467](https://github.com/elastic/quark/pull/250/commits/41874678e53e0fafa02b96517324652843ecea4f)

haesbaert (approved):
Does this work? I'm writing this from `gh pr review`.
I'm trying to approve, nice work!

Issue #251, opened 2025-10-28
[BUG] bugs found with bugfinder


stanek-michal:
1)

https://github.com/elastic/quark/blob/main/bpf_queue.c#L566
Bugfinder found that the `for` loop fails on one edge case - where the `cgroup2` line happens to be the first line in /proc/mounts.
In that case, the code lies to you because `line` variable would contain just the first word of the line, not the actual full line.
This stops being the case if `cgroup2` is on any other lines, because then the `strtok_r` with \n as separator returns an actual full line which we can then `strchr` through with ' ' separator.

Tricky!

2)
https://github.com/elastic/quark/blob/main/bpf_queue.c#L604
in any of these three `goto fail;`, we forget to remove the directory created in `mkdtemp`. The `cgroup2_umount_tmp` call will do this cleanup but only when `umount_path` is set, which it isn't for the three goto's.

3)
https://github.com/elastic/quark/blob/main/bpf_queue.c#L1028
bpf_queue_update_stats() iterates over num_possible_cpus and does `=` instead of `+=` so it's not accumulating lost counters from all cpus but just from the last cpu

**plus some lower severity bugs below:**

4)
https://github.com/elastic/quark/blob/main/bpf_queue.c#L232
this one is not a big deal because we have bigger problems if `malloc` returns NULL, but here if it does, it will leave a weird structure where `raw->exec.ext.args_len` is non zero but raw->exec.ext.args is empty. we should at least clear `args_len` in the NULL case

5) 
https://github.com/elastic/quark/blob/main/quark.c#L3519
low severity cause again relies on `strdup` failing but if it does here, `si->ip_addrs` is an invalid dangling pointer that gets used on next loop iteration

6) 
https://github.com/elastic/quark/blob/main/quark.c#L248
`raw_event_target_age` could use a guard in the end to make sure `v` isn't negative (looks like it could happen depending on the length settings) and also to ensure we don't divide by 0

7)
https://github.com/elastic/quark/blob/main/quark.c#L2626
end of `sproc_stat` - looks like we're hardcoding hz=100 here?

haesbaert:
This is awesome!

I've confirmed all except 4., but it should still be sanitized, thanks a lot for finding these!

PR #252, opened 2025-10-28
Fix edge case in cgroup2 parsing

stanek-michal:
Fix typo that made the code fail in a case where `cgroup2` is the first line of `/proc/mounts`


haesbaert (approved):
BAAAAAAAD haesbaert, thanks for finding it and fixing !
I loves it.

haesbaert:
Issue #251

PR #283, opened 2025-11-20
Preliminary Rule System

haesbaert:
This implements the barebones of a rule system for filtering events.

    A ruleset is a collection of rules.
    A rule has zero or more rule_fields. Each rule_field expresses one attribute
    that must match, for example process.pid is one field, and file.path is another.
    A rule matches when _all_ its fields match.
    A rule with no fields always matches (can be a catch-all as the last rule).
    A ruleset matches the first rule it matches, meaning this is a match-first scheme.

    Currently we have only two actions RA_PASS and RA_DROP, if nothinig matches,
    it's PASS.
    For now, only 4 rule_fields, process.pid, process.ppid, process.filename and
    file.path.

    The next step is to design a proper DSL ala pf.conf(5), that then compiles an
    ascii line-based ruleset into a quark_ruleset.

Note that two minor changes were piggy backed in different commits (gc_collect + re-align)

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -4055,6 +4055,7 @@ quark_queue_default_attr(struct quark_queue_attr *qa)
 	qa->hold_time = 1000;		/* one second */
 	qa->max_env = 4096;		/* one page per process */
 	qa->kubefd = -1;		/* disabled */
+	qa->ruleset = NULL;		/* no rules */
 }
 
 int
@@ -4126,6 +4127,7 @@ quark_queue_open(struct quark_queue *qq, struct quark_queue_attr *qa)
 	qq->cache_grace_time = MS_TO_NS(qa->cache_grace_time);
 	qq->hold_time = qa->hold_time;
 	qq->max_env = qa->max_env;
+	qq->ruleset = qa->ruleset;
 	qq->length = 0;
 	qq->epollfd = -1;
 	if (qq->flags & QQ_MIN_AGG)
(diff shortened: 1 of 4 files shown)


quark.c
```diff
@@ -4736,6 +4738,217 @@ raw_event_tty(struct quark_queue *qq, struct raw_event *raw)
...
+			return (qev->process->proc_ppid == field->pid);
+		break;
+	case RF_PROCESS_FILENAME:
+		if (qev->process != NULL &&
+		    (qev->process->flags & QUARK_F_FILENAME) &&
+		    qev->process->filename != NULL)
+			return (path_match(field, qev->process->filename));
+		break;
+	case RF_FILE_PATH:
+		if (qev->file != NULL)
```
nicholasberlin: I notice that you checked `qev->process->filename != NULL`, why not `qev->file->path != NULL` here?

quark.c
```diff
@@ -4736,6 +4738,217 @@ raw_event_tty(struct quark_queue *qq, struct raw_event *raw)
...
+}
+
+static int
+quark_rule_append_any_path(struct quark_rule *rule, const char *path,
+    enum rule_field_code code)
+{
+	struct rule_field	 rf;
+	char			*p;
+
+	if (strlen(path) >= PATH_MAX)
```
nicholasberlin: should evaluate against `sizeof(rf.path)` in case we make it smaller, which we probably won't, but still.

haesbaert: I did this intentionally cause I want to make rf.path a pointer and allocate it, so I'd leave it like this so I don't have to remember to change it.

quark.c
```diff
@@ -4761,10 +4974,19 @@ quark_queue_get_event(struct quark_queue *qq)
...
 	struct raw_event	*raw;
 	struct quark_event	*qev;
+	struct quark_rule	*rule;
 
+	/* GC all processes and sockets that exited after some grace time */
+	gc_collect(qq);
+
+	/* GC all processes and sockets that exited after some grace time */
+	gc_collect(qq);
+
```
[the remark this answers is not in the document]
haesbaert: woooooooooooooooooooooooooooooooooooooooooooooooooooops.
that was me screwing the rebases, great catch!

quark.c
haesbaert: basically file->path cannot be null, it points to itself

PR #298, opened 2026-02-06
Fix qreadlinkat ret val and truncation detection

nicholasberlin:
## Fix qreadlinkat return value and truncation detection

`qreadlinkat()` had two bugs:

1. It returned `strlen(pathname)` (the pathname argument, e.g. `3` for `"exe"`) instead of the length of the resolved symlink target. This meant callers always got the wrong value back.

2. Truncation was undetectable. Since `readlinkat(2)` was called with `bufsiz - 1`, callers checking `n >= bufsiz` could never see truncation. Fixed by bumping `n` to `bufsiz` when `readlinkat(2)` fills the entire buffer (`n == bufsiz - 1`).

### Changes

- **`qutil.c`**: Return `n` instead of `strlen(pathname)`. When `readlinkat(2)` returns `bufsiz - 1` (buffer full), increment `n` to `bufsiz` so callers can detect truncation via `n >= bufsiz`.

### DIFF
--- a/qutil.c
+++ b/qutil.c
@@ -52,9 +52,9 @@ qwrite(int fd, const void *buf, size_t count)
 }
 
 /*
- * Safer readlinkat(2), guarantees termination and returns strlen(pathname) so
- * caller can check truncation, like strlcpy(). Compare to >= 0 if truncation is
- * acceptable.
+ * Safer readlinkat(2), guarantees termination and returns the number of bytes
+ * placed in buf (excluding NUL) so caller can check truncation, like strlcpy().
+ * If the return value is >= bufsiz, truncation occurred.
  */
 ssize_t
 qreadlinkat(int dfd, const char *pathname, char *buf, size_t bufsiz)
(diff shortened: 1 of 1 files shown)


nicholasberlin:
This bug was found and coded by the bot: claude-4.6-opus-max

quark-test.c
```diff
@@ -2099,6 +2099,62 @@ t_rule_parser(const struct test *t, struct quark_queue_attr *qa)
...
+		/* buf must still be terminated */
+		assert(small[sizeof(small) - 1] == 0 ||
+		    strlen(small) < sizeof(small));
+	}
+
+	close(dfd);
+
+	return (0);
+}
+
```
haesbaert: I don't think we need 150 lines to test a basic function, I'd zap this

qutil.c
```diff
@@ -63,11 +63,15 @@ qreadlinkat(int dfd, const char *pathname, char *buf, size_t bufsiz)
 
 	if (bufsiz < 2)
 		return (errno = EINVAL, -1);
-	if ((n = readlinkat(dfd, pathname, buf, bufsiz - 1)) == -1)
```
haesbaert: how about something like:
```
ssize_t
qreadlinkat(int dfd, const char *pathname, char *buf, size_t bufsiz)
{
	ssize_t n;

	if (bufsiz < 2)
		return (errno = EINVAL, -1);
	if ((n = readlinkat(dfd, pathname, buf, bufsiz - 1)) == -1)
		return (-1);
	buf[n] = 0;
	if (n == (bufsiz - 1))
		n++;

	return (n);
}
```
It's one extra if and changing the return to `n`

haesbaert (commented):
Thanks for findings this!
I don't think we need a test, it adds too much code for somethign small, I've added a different version of the function, leme know what you think

haesbaert (approved):

PR #305, opened 2026-03-30
Cope with PREEMPT_LAZY in our bpf probes.

haesbaert:
Linux introduced a new form of preemption which is to become the default, PREEMPT_LAZY was introduced around 6.13 in 7c70cb94d29cd325fabe4a818c18613e3b9919a1.

It is a combination of voluntary and realtime preemption schemes. When one thread is to be preempted, it gets asked nicely by havin a flag set, it then voluntary preempts if it reaches a preemption point. But, if the scheduler tries to set the flag, but the flag is already set, meaning the thread has been a bad boi and didn't voluntary preempt itself, it forces a full switch on the spot.

This causes our probe to become actually preemptive, which is a first, and all our code has been written assuming probes would never be preemptive.

Around 6.10 bpf_{disable,enable}_preemption() has been added as a KFUNC, this commit adds it around every probe we install it if the KFUNC is there. This disregards the actual preemption model of the kernel, as we can't check it during runtime (it can change at any time). As a bonus, we should now actually function in RT kernels (I guess?).

The observed behaviour was a nasty corruption of the event buffer, which then made it to the ring buffer with invalid offsets, the "fields" and "field" size would get corrupted, as can be seen in this partial analysis of the ring buffer:

```
(gdb) x/256bx &$exec->vl_fields
0x7ffff6d99c00: 0x23 0x00 0x00 0x00 0x00 0x00 0x00 0x00 (0x23 should be nfields (4 bytes)) WHERE IS NFIELDS? WHERE IS SIZE?
0x7ffff6d99c08: 0x08 0x00 0x00 0x00 0x1b 0x00 0x00 0x00 (0x08 is in the upper half of the size_t (8bytes), 34359738368) 8 is also=EBPF_VL_FIELD_PIDS_SS_CGROUP_PATH
0x7ffff6d99c10: 0x2f 0x73 0x79 0x73 0x74 0x65 0x6d 0x2e (/system.slice/nrpe.service
0x7ffff6d99c18: 0x73 0x6c 0x69 0x63 0x65 0x2f 0x6e 0x72 .....
0x7ffff6d99c20: 0x70 0x65 0x2e 0x73 0x65 0x72 0x76 0x69 ....
0x7ffff6d99c28: 0x63 0x65 0x00 0x00 0x03 0xd0 0x16 0x00 ....)
```

This was figured out by Nick Berlin, Yours Truly and the help of a $VERY_NICE_AND_HELPFUL_CUSTOMER (Many Thanks!)

We could try to go over all data structures of our probes and think about re-entrancy, but I claim it to be the wrong approach, if all was written with the mindset of no preemption, the safest approach should be to recreate that view, and not try to work around it with it.

Nick provided a test program that rapidly reproduces the problem making quark-mon(8) crash, it should be visible in this commit's PR, and we will incorporate it in testing somehow.

### DIFF
--- a/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
@@ -59,13 +59,25 @@ SEC("fexit/inet_csk_accept")
 int BPF_PROG(fexit__inet_csk_accept)
 {
     struct sock *ret = FUNC_RET_READ(___type(ret), inet_csk_accept);
-    return inet_csk_accept__exit(ret);
+    int r;
+
+    preempt_disable();
+    r = inet_csk_accept__exit(ret);
+    preempt_enable();
+
+    return r;
 }
 
 SEC("kretprobe/inet_csk_accept")
 int BPF_KRETPROBE(kretprobe__inet_csk_accept, struct sock *ret)
 {
-    return inet_csk_accept__exit(ret);
+    int r;
+
+    preempt_disable();
+    r = inet_csk_accept__exit(ret);
+    preempt_enable();
+
+    return r;
 }
 
 static int tcp_connect(struct sock *sk, int ret)
@@ -98,7 +110,13 @@ out:
 SEC("fexit/tcp_v4_connect")
 int BPF_PROG(fexit__tcp_v4_connect, struct sock *sk, struct sockaddr *uaddr, int addr_len, int ret)
 {
-    return tcp_connect(sk, ret);
+    int r;
+
+    preempt_disable();
+    r = tcp_connect(sk, ret);
+    preempt_enable();
+
+    return r;
 }
 
 SEC("kprobe/tcp_v4_connect")
@@ -106,9 +124,14 @@ int BPF_KPROBE(kprobe__tcp_v4_connect, struct sock *sk)
 {
     struct ebpf_events_state state = {};
     state.tcp_v4_connect.sk        = sk;
+
+    preempt_disable();
     if (ebpf_events_is_trusted_pid())
-        return 0;
+        goto out;
     ebpf_events_state__set(EBPF_EVENTS_STATE_TCP_V4_CONNECT, &state);
+
+out:
+    preempt_enable();
     return 0;
 }
 
@@ -116,18 +139,31 @@ SEC("kretprobe/tcp_v4_connect")
 int BPF_KRETPROBE(kretprobe__tcp_v4_connect, int ret)
 {
     struct ebpf_events_state *state;
+    int r;
 
+    preempt_disable();
+    r = 0;
     state = ebpf_events_state__get(EBPF_EVENTS_STATE_TCP_V4_CONNECT);
     if (!state)
-        return 0;
+        goto out;
 
-    return tcp_connect(state->tcp_v4_connect.sk, ret);
+    r = tcp_connect(state->tcp_v4_connect.sk, ret);
+
+out:
+    preempt_enable();
+    return r;
 }
 
 SEC("fexit/tcp_v6_connect")
 int BPF_PROG(fexit__tcp_v6_connect, struct sock *sk, struct sockaddr *uaddr, int addr_len, int ret)
 {
-    return tcp_connect(sk, ret);
+    int r;
+
+    preempt_disable();
+    r = tcp_connect(sk, ret);
+    preempt_enable();
+
+    return r;
 }
 
 SEC("kprobe/tcp_v6_connect")
@@ -135,9 +171,14 @@ int BPF_KPROBE(kprobe__tcp_v6_connect, struct sock *sk)
 {
     struct ebpf_events_state state = {};
     state.tcp_v6_connect.sk        = sk;
+
+    preempt_disable();
     if (ebpf_events_is_trusted_pid())
-        return 0;
+        goto out;
     ebpf_events_state__set(EBPF_EVENTS_STATE_TCP_V6_CONNECT, &state);
+
+out:
+    preempt_enable();
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


haesbaert:
Program for reproducing the bug:
```
// SPDX-License-Identifier: Apache-2.0
// Reproducer for per-CPU event buffer race under PREEMPT_LAZY.
//
// The bug: bpf_ringbuf_output() reads from a per-CPU buffer while another BPF
// program (preempted onto the same CPU) overwrites it, creating a chimera
// event in the ringbuf. Under PREEMPT_LAZY, normal CFS scheduling defers
// preemption (TIF_NEED_RESCHED_LAZY), so BPF programs usually complete before
// being preempted. But TIF_NEED_RESCHED (immediate) is set by signal delivery,
// high-priority wakeups, etc. — causing preemption mid-BPF-helper.
//
// Strategy: pin fork+exec+exit workers to one CPU and periodically send
// signals to create bursts of TIF_NEED_RESCHED, maximizing the chance of
// preempting bpf_ringbuf_output's memcpy.
//
// Usage:
//   gcc -O2 -o repro-buffer-race repro-buffer-race.c -lpthread
//   echo lazy | sudo tee /sys/kernel/debug/sched/preempt
//   # Terminal 1: sudo ./quark-mon
//   # Terminal 2: sudo ./repro-buffer-race [cpu] [seconds] [nworkers]

#define _GNU_SOURCE
#include <errno.h>
#include <pthread.h>
#include <sched.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <time.h>
#include <unistd.h>

static volatile sig_atomic_t stop;

static void
sighandler(int sig)
{
	(void)sig;
	stop = 1;
}

static void
nop_handler(int sig)
{
	(void)sig;
}

static void
pin_to_cpu(int cpu)
{
	cpu_set_t set;
	CPU_ZERO(&set);
	CPU_SET(cpu, &set);
	if (sched_setaffinity(0, sizeof(set), &set) == -1) {
		perror("sched_setaffinity");
		_exit(1);
	}
}

struct worker_args {
	int cpu;
	int id;
	unsigned long count;
};

static volatile pid_t child_pids[64];
static volatile int nchildren;

static void *
exec_exit_worker(void *arg)
{
	struct worker_args *wa = arg;
	int slot = wa->id;

	pin_to_cpu(wa->cpu);

	while (!stop) {
		pid_t p = fork();
		if (p == 0) {
			execl("/bin/true", "true", NULL);
			_exit(1);
		}
		if (p > 0) {
			child_pids[slot] = p;
			waitpid(p, NULL, 0);
			child_pids[slot] = 0;
		}
		wa->count++;
	}

	return NULL;
}

// Sends signals to worker children to create TIF_NEED_RESCHED bursts.
// When a signal is delivered to a process, the scheduler sets the immediate
// (non-lazy) TIF_NEED_RESCHED flag on the currently running task, forcing
// preemption at the next interrupt return — even inside bpf_ringbuf_output.
static void *
signal_bombarder(void *arg)
{
	struct worker_args *wa = arg;
	int nworkers = wa->id;
	struct timespec ts = { .tv_sec = 0, .tv_nsec = 100000 }; // 100us

	pin_to_cpu(wa->cpu);

	while (!stop) {
		for (int i = 0; i < nworkers; i++) {
			pid_t p = child_pids[i];
			if (p > 0)
				kill(p, SIGUSR1);
		}
		nanosleep(&ts, NULL);
		wa->count++;
	}

	return NULL;
}

// Yield thread to add scheduling pressure
static void *
yield_worker(void *arg)
{
	struct worker_args *wa = arg;
	pin_to_cpu(wa->cpu);

	while (!stop) {
		sched_yield();
		wa->count++;
	}

	return NULL;
}

int
main(int argc, char **argv)
{
	int cpu = 0;
	int duration = 30;
	int nworkers = 4;

	if (argc > 1)
		cpu = atoi(argv[1]);
	if (argc > 2)
		duration = atoi(argv[2]);
	if (argc > 3)
		nworkers = atoi(argv[3]);
	if (nworkers > 64)
		nworkers = 64;

	int total_threads = nworkers + nworkers + 1;
	pthread_t *threads = calloc(total_threads, sizeof(*threads));
	struct worker_args *args = calloc(total_threads, sizeof(*args));
	if (!threads || !args) {
		perror("calloc");
		return 1;
	}

	fprintf(stderr,
	    "Reproducing bpf_ringbuf_output preemption race on CPU %d\n"
	    "  %d exec/exit workers + %d yield workers + 1 signal bombarder\n"
	    "  Duration: %d seconds\n"
	    "  Run quark-mon in another terminal and watch for a crash.\n\n",
	    cpu, nworkers, nworkers, duration);

	signal(SIGALRM, sighandler);
	signal(SIGINT, sighandler);
	signal(SIGUSR1, nop_handler);
	alarm(duration);

	pin_to_cpu(cpu);
	memset((void *)child_pids, 0, sizeof(child_pids));

	int t = 0;

	for (int i = 0; i < nworkers; i++) {
		args[t].cpu = cpu;
		args[t].id = i;
		args[t].count = 0;
		pthread_create(&threads[t], NULL, exec_exit_worker, &args[t]);
		t++;
	}

	for (int i = 0; i < nworkers; i++) {
		args[t].cpu = cpu;
		args[t].id = i;
		args[t].count = 0;
		pthread_create(&threads[t], NULL, yield_worker, &args[t]);
		t++;
	}

	// Signal bombarder — sends SIGUSR1 to children to force
	// TIF_NEED_RESCHED (immediate) on whatever is running on this CPU
	args[t].cpu = cpu;
	args[t].id = nworkers;
	args[t].count = 0;
	pthread_create(&threads[t], NULL, signal_bombarder, &args[t]);
	t++;

	for (int i = 0; i < t; i++)
		pthread_join(threads[i], NULL);

	unsigned long total_execs = 0, total_yields = 0;
	for (int i = 0; i < nworkers; i++)
		total_execs += args[i].count;
	for (int i = nworkers; i < nworkers * 2; i++)
		total_yields += args[i].count;

	fprintf(stderr, "\nDone: %lu exec/exit cycles, %lu yields, %lu signal rounds\n",
	    total_execs, total_yields, args[t - 1].count);

	free(threads);
	free(args);

	return 0;
}
```

elastic-ebpf/GPL/Events/Network/Probe.bpf.c
```diff
@@ -384,25 +457,49 @@ int sk_maybe_save_tgid(struct bpf_sock *sk)
 SEC("cgroup/sendmsg4")
 int sendmsg4(struct bpf_sock_addr *sa)
 {
-    return sk_maybe_save_tgid(sa->sk);
+    int r;
+
+    preempt_disable();
+    r = sk_maybe_save_tgid(sa->sk);
```
nicholasberlin, line 463: make `sk_maybe_save_tgid` `static` too?

haesbaert: good point, done!

nicholasberlin (approved):

PR #307, opened 2026-04-13
Add skeleton for the nova bpf probes

haesbaert:
Three commits, two minor nits, main one is:
```
Date:   Mon Apr 13 10:48:36 2026 +0200

    Add skeleton for the nova bpf probes

    It's here, we're writing new probes. The goal is to be able to express the
    ruleset inside bpf, which preliminary results show it shouldn't be too hard.

    This will be a new queue QQ_NOVA, so that we can develop and break things freely
    while keeping QQ_BPF as it is, for now this is the plan, we can re-evaluate it
    in the medium term.

    Naming is hard, initially it would be qbpf, but it makes it too easy to mix
    things up, for example qbpf_queue.c X bpf_queue.c, QQ_EBPF vs QQ_QBPF.

    So pretend we're cool and name it "nova", no way to mix things up there.

    This only hooks up the build system with empty declarations.
```

### DIFF
--- a/nova.bpf.c
+++ b/nova.bpf.c
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0-only OR BSD-2-Clause
+/* Copyright (c) 2026 Elastic NV */
+
+#include "vmlinux.h"		/* XXX still getting the old one XXX */
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#include <linux/limits.h>
+
+SEC("lsm/file_open")
+int BPF_PROG(test_path, struct file *file)
+{
+	return (0);
+}
+
+char _license[] SEC("license") = "Dual BSD/GPL";
--- a/Makefile
+++ b/Makefile
@@ -184,11 +184,10 @@ ifdef CENTOS7
 LIBBPF_EXTRA_CFLAGS+= -Wno-address
 endif
 
-# BPFPROG (kernel side)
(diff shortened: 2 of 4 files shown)


nova.bpf.c
nicholasberlin, line 1: I see this in the soon-to-be legacy *.probe.c files:
```
// SPDX-License-Identifier: GPL-2.0-only OR BSD-2-Clause

/*
 * Copyright (C) 2021 Elasticsearch BV
 *
 * This software is dual-licensed under the BSD 2-Clause and GPL v2 licenses.
 * You may choose either one of them if you use this software.
 */
```

Should that be in here?

nova.bpf.c
```diff
@@ -0,0 +1,17 @@
...
+
+#include <errno.h>
+
+SEC("lsm/file_open")
+int BPF_PROG(test_path, struct file *file)
+{
+	return (0);
+}
+
+char _license[] SEC("license") = "Dual BSD/GPL";
```
nicholasberlin, line 18: I don't see this in the *.probe.c files. Do you know how it's added for them?

nicholasberlin (approved):

nova.bpf.c
haesbaert: char LICENSE[] SEC("license") = "Dual BSD/GPL";
^^ this is in bpf_probes.c

nova.bpf.c
haesbaert: oh yeah I should copyright nova.bpf.c, well spot

nicholasberlin (approved):
Changes LGTM

PR #309, opened 2026-04-16
Nova ruleset evaluation in bpf probes

haesbaert:
This implements the basic rule matching for nova probes, this diff does
basically two things:

o - Actually load QQ_NOVA probes
o - Installs the quark_ruleset into the BPF probes so we can evaluate rules in
    kernel land. Path matching is still missing but this is a good start.

We support 1024 rules, and for the first version we will require 5.17 so we have
bpf_loop(). We do some slightly smart things like diminishing the amount of
branches the verifier has to calculate by doing branchless comparisons in the
evaluation loop, this is likely not necessary for 5.17 as bpf_loop() saves a lot
of budget.

Next step is implementing path matching via bpf_strncmp() or manually, since we
likely don't need TRIE for this.

### DIFF
--- a/nova.bpf.c
+++ b/nova.bpf.c
@@ -7,11 +7,112 @@
 #include <bpf/bpf_tracing.h>
 #include <bpf/bpf_core_read.h>
 
-#include <linux/limits.h>
+#include "nova.h"
+
+/* XXX check if there isn't something like this on the headers XXX */
+#define LOOP_CONTINUE	0
+#define LOOP_STOP	1
+
+#define E2BIG		7
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, struct nova_rule);
+} scratch_rule SEC(".maps");
+
+#define my_rule_eval_ctx()	bpf_map_lookup_elem(&scratch_rule, &(int){0})
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, __u32);
+	__type(value, struct nova_rule);
+	__uint(max_entries, NOVA_MAX_RULES);
+} ruleset SEC(".maps");
+
+struct rule_eval_loop_ctx {
+	struct nova_rule	*rctx;
+	struct nova_rule	*match;
+};
+
+extern void bpf_preempt_disable(void) __ksym __weak;
+
+static void preempt_disable(void)
+{
+	if (bpf_ksym_exists(bpf_preempt_disable))
+		bpf_preempt_disable();
+}
+
+extern void bpf_preempt_enable(void) __ksym __weak;
+
+static void preempt_enable(void)
+{
+	if (bpf_ksym_exists(bpf_preempt_enable))
+		bpf_preempt_enable();
+}
+
+static int
+rule_eval_loop(__u32 i, struct rule_eval_loop_ctx *lctx)
+{
+	struct nova_rule	*cand;
+	struct nova_rule	*rctx;
+
+	rctx = lctx->rctx;
+	if (rctx == NULL ||
+	    ((cand = bpf_map_lookup_elem(&ruleset, &i)) == NULL))
+		return (LOOP_CONTINUE); /* XXX signal something is really wrong XXX */
+	rctx->fields = 0;
+	rctx->fields |= (__u64)(rctx->pid == cand->pid) * QUARK_RF_PROCESS_PID;
+	rctx->fields |= (__u64)(rctx->ppid == cand->ppid) * QUARK_RF_PROCESS_PPID;
+	rctx->fields |= (__u64)(rctx->uid == cand->uid) * QUARK_RF_PROCESS_UID;
+	rctx->fields |= (__u64)(rctx->gid == cand->gid) * QUARK_RF_PROCESS_GID;
+	rctx->fields |= (__u64)(rctx->sid == cand->sid) * QUARK_RF_PROCESS_SID;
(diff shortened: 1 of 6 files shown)


nova.bpf.c
```diff
@@ -7,11 +7,92 @@
...
+	if (lctx.match == NULL)
+		return (0);
+
+	return (0);
+}
+
 SEC("lsm/task_alloc")
 int BPF_PROG(task_alloc, struct task_struct *task, __u64 clone_flags)
 {
+	struct nova_rule	*rctx = my_rule_eval_ctx();
```
nicholasberlin: disable preemption?

haesbaert: ah yeah, better add it now least I forget, will do it.

nova_queue.c
```diff
@@ -34,6 +34,94 @@ struct quark_queue_ops queue_ops_nova = {
...
+			break;
+		default:
+			errno = EINVAL;
+			qwarn("bad field->code %llu", field->code);
+			return (-1);
+			break;
+		}
+	}
+
+	nr->fields = (__u64)qr->fields;
```
nicholasberlin: the bot says this:
 `nr->fields = (__u64)qr->fields;` stores a pointer in the bitmask (`nova_queue.c:90`)
  quark_rule.fields is a struct quark_rule_field * (array pointer); nova_rule.fields is the match bitmask the BPF loop tests via (rctx->fields & cand->fields) == cand->fields. This assignment casts a userspace pointer into the
  bitmask, so matching will (essentially) never fire — and if it did, for the wrong reason. The fix is to OR field->code (now bit flags) across qr->fields[0..n_fields-1]. Consider renaming one of the two fields members to avoid
   this collision.

nova_queue.c
```diff
@@ -34,6 +34,94 @@ struct quark_queue_ops queue_ops_nova = {
...
+static int
+load_ruleset(struct quark_queue *qq, struct nova_queue *nqq)
+{
+	u32			 i;
+	struct nova_bpf		*nova_bpf;
+	struct nova_rule	 nr;
+
+	if (qq->ruleset == NULL)
+		return (0);
+	if (qq->ruleset->n_rules >= NOVA_MAX_RULES)
```
nicholasberlin: I think this should be `>`, or else 1024 will rejected.

nova.bpf.c
```diff
@@ -7,11 +7,92 @@
...
+{
+	struct nova_rule	*cand;
+	struct nova_rule	*rctx;
+
+	rctx = lctx->rctx;
+	if (rctx == NULL ||
+	    ((cand = bpf_map_lookup_elem(&ruleset, &i)) == NULL))
+		return (LOOP_CONTINUE); /* XXX signal something is really wrong XXX */
+	rctx->fields = 0;
+	rctx->fields |= (__u64)(rctx->pid == cand->pid) * QUARK_RF_PROCESS_PID;
```
nicholasberlin, line 66: nifty, multiplying by 0 or 1.

nova.bpf.c
```diff
@@ -7,11 +7,92 @@
...
+	lctx.rctx = rctx;
+	lctx.match = NULL;
+
+	r = bpf_loop(NOVA_MAX_RULES, rule_eval_loop, &lctx, 0);
+	if (r < 0) {
+		bpf_printk("bad bpf_loop %d", r);
+		return (0);
+	}
+	if (lctx.match == NULL)
+		return (0);
```
nicholasberlin, line 97: clearly this is work-in-progess, but return 1 here?

nova_queue.c
haesbaert: bleaaaaaah, my bad.

nova_queue.c
haesbaert: ack

nova.bpf.c
haesbaert: I'll change this a lot so I wouldn't mind it now

nova.bpf.c
```diff
@@ -7,11 +7,92 @@
...
+	return (0);
+}
+
 SEC("lsm/task_alloc")
 int BPF_PROG(task_alloc, struct task_struct *task, __u64 clone_flags)
 {
+	struct nova_rule	*rctx = my_rule_eval_ctx();
+	int			 r;
+
+	r = rule_eval(rctx);
```
nicholasberlin, line 111: so, eventually, `rctx->pid`, etc will be populated in `rule_eval`?

nova_queue.c
```diff
@@ -34,6 +34,94 @@ struct quark_queue_ops queue_ops_nova = {
...
+		case QUARK_RF_PROCESS_FILENAME:
+			/* TODO */
+			errno = ENOTSUP;
+			qwarn("QUARK_RF_PROCESS_FILENAME");
+			return (-1);
+			break;
+		case QUARK_RF_FILE_PATH:
+			/* TODO */
+			errno = ENOTSUP;
+			qwarn("QUARK_RF_PROCESS_FILENAME");
```
nicholasberlin:
```suggestion
			qwarn("QUARK_RF_FILE_PATH");
```

no big deal, this will clearly be removed soon.

nova.bpf.c
haesbaert: correct, since this is supposed to be generic, as in run from different types of probes/attach points, the caller will fill in the details and call the val function

nova_queue.c
haesbaert: nice catch!

haesbaert:
so canonical is robbing us of our productivity, there's something going on with the apt repositories, that's why CI is failing

nicholasberlin (approved):

PR #311, opened 2026-04-23
Minor fixes, Update Container & clang-20+ fixes

haesbaert:
Once upon a time this was a very simple group of commits, now it's a very simple group of commits + 1 not simple.

Each commit is independent:

* Minor style nits - easy peasy
* Fix golang builds on centos7 - easy peasy
* Improve t_shm_open() - easy peasy
* Update the docker builder to ubuntu26.04 - easy peasy
* Make sure our probes load on 5.10 and clang-20+ - **oh noes**

Here is the last commit log:
```
commit e47c20fbc10a9da4f433799f7a8c3a42a16fdc57 (HEAD -> cut-the-slack, origin/cut-the-slack)
Date:   Fri Apr 24 11:20:56 2026 +0200

    Make sure our probes load on 5.10 and clang-20+

    5.10 and older backported kernels, like 4.18 will lose bound tracking of a
    register when it narrows from 64bit to 32bit, this makes the verifier freak out
    depending on how clang generates the code.

    We were lucky that clang-{18,19} emitted code the verifier was happy on 5.10.
    But on clang-{20,21} it will emit code the verifier complains about unbounded
    check. In the DNS case I had to add a barrier to force clang to not optimize the
    recomputation away. The others were easier and we fix it by just making sure we
    stay always on 32bit.

    Just to be clear, this is an issue on older (5.10ish) and depends on the emitted
    code.
```

### DIFF
--- a/elastic-ebpf/GPL/Events/Helpers.h
+++ b/elastic-ebpf/GPL/Events/Helpers.h
@@ -155,20 +155,36 @@ static long read_kernel_str_or_empty_str(void *dst, int size, const void *unsafe
     return ret;
 }
 
-static long ebpf_argv__fill(char *buf, size_t buf_size, const struct task_struct *task)
+// These should be kept _well_ under half the size of the event_buffer_map or
+// the verifier will be unhappy due to bounds checks. Putting a cap on these
+// things also prevents any one process from DoS'ing and filling up the
+// ringbuffer with super rapid-fire events.
+#define ARGV_MAX 20480u
+#define ENV_MAX 40960u
+
+static long ebpf_argv__fill(char *buf, const struct task_struct *task)
 {
-    unsigned long start, end, size;
+    unsigned long start, end;
+    u32 size;
 
     start = BPF_CORE_READ(task, mm, arg_start);
     end   = BPF_CORE_READ(task, mm, arg_end);
 
-    if (end <= start) {
-        buf[0] = '\0';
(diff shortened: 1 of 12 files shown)


nfritts (approved):

elastic-ebpf/GPL/Events/Helpers.h
```diff
@@ -155,20 +155,33 @@ static long read_kernel_str_or_empty_str(void *dst, int size, const void *unsafe
...
+static long ebpf_argv__fill(char *buf, const struct task_struct *task)
 {
-    unsigned long start, end, size;
+    unsigned long start, end;
+    u32 size;
 
     start = BPF_CORE_READ(task, mm, arg_start);
     end   = BPF_CORE_READ(task, mm, arg_end);
 
-    if (end <= start) {
```
nicholasberlin, line 165: I'm guessing `end` being less than `start` doesn't happen in real life.

Could put `if (end <= start) return 1;` before the subtraction. 🤷

nicholasberlin (approved):
One nitpick, but LGTM. Nice work tracking down the verifier issues.

elastic-ebpf/GPL/Events/Helpers.h
haesbaert: I've just pushed a version with that, will commit once the test passes

PR #314, opened 2026-04-27
Rename a bunch of fields

haesbaert:
Cut the _PROCESS_ from QUARK_RF_*, it's just bloat and PROCESS is too long, it
makes lines expand too much.

Also rename filename to exe, this is a breaking change.
Filename was inherited from the kprobe backend, where the kprobe itself calls it
filename, I can't stand it. EXE is more descriptive.


nicholasberlin (approved):

ricardoamaro (approved):
LGTM! The renames look clean and applied consistently across all 8 files.

haesbaert:
Issue #315

Issue #315, opened 2026-04-27
Add LSM Probes (QQ_NOVA)


haesbaert:

haesbaert:
https://github.com/elastic/quark/pull/313 https://github.com/elastic/quark/pull/314 https://github.com/elastic/quark/pull/315 https://github.com/elastic/quark/pull/316 https://github.com/elastic/quark/pull/317 https://github.com/elastic/quark/pull/318

ricardoamaro:
Here goes a tentative description:

Add BPF LSM-based in-kernel rule evaluation (QQ_NOVA) to quark. Match on the
path being executed with LPM + single-`*` infix globs, drop non-matching events
in the kernel, expose per-rule stats.

## Work

- [x] Per-CPU eval/match stats; guard `libbpf_num_possible_cpus()` ([#313](https://github.com/elastic/quark/pull/313))
- [x] Rename `QUARK_RF_PROCESS_*` → `QUARK_RF_*`, `filename` → `exe` — breaking ([#314](https://github.com/elastic/quark/pull/314))
- [x] `PATH_LPM_PREFIXLEN()` via `bpf_d_path()`; hook the path being executed, not the calling task's exe ([#316](https://github.com/elastic/quark/pull/316))
- [x] Infix path grammar with a single `*`, no backtracking ([#318](https://github.com/elastic/quark/pull/318))
- [x] Wire infix matching into the QQ_NOVA BPF program (follow-up noted in [#318](https://github.com/elastic/quark/pull/318): "Next diffs will be adding this to QQ_NOVA")
- [ ] Match on both calling-task exe and target path (follow-up noted in [#316](https://github.com/elastic/quark/pull/316): "we will likely want to match on both")
- [ ] Userland API / docs for QQ_NOVA rule fields (no PR yet)
- [ ] Tests: rule poisoning, counters, path-match edge cases (poisoning-counter reset landed in [#318](https://github.com/elastic/quark/pull/318); broader coverage TBD)

## Notes

- `quark_rule_field{}` grew 16 → 32 bytes per path field to carry the wildcard descriptor ([#318](https://github.com/elastic/quark/pull/318)).

ricardoamaro:
Synced with @haesbaert today. Deferring the remaining work to sensor iteration 2 (post-September 2026).

The blocking use case this was built for belongs in the new sensor, and iteration 1 ships without inline blocking. No point spending capacity here now without that context.

**What's done:** #313, #314, #316, #318 merged.

[...]

PR #316, opened 2026-04-27
Improve prefixlen calculation and fix executable matching. Issue 315

haesbaert:
This improves prefixlen calculation by providing a macro PATH_LPM_PREFIXLEN().
bpf_d_path() returns the length with NUL included, which is exactly what we
want, prefixlen for a path matching is basically (sizeof(meta) + path + NUL), in
bits, and clampped to PATH_LPM_KEYLEN.

Also, instead of hooking in task_alloc, hook in the actual executable path,
before we were matching on the current task executable, instead of the path
trying to be executed. This will likely change in the future, since we will
likely want to match on both.

While here, add some more silly cases to the test, so we can try bash vs bc for
instance.

issue #315

### DIFF
--- a/nova.bpf.c
+++ b/nova.bpf.c
@@ -18,6 +18,16 @@
 
 #define E2BIG		7
 
+#define MAX(_a, _b)	((_a) > (_b) ? (_a) : (_b))
+#define MIN(_a, _b)	((_a) < (_b) ? (_a) : (_b))
+
+#define PATH_LPM_KEYLEN							\
+	(sizeof(struct path_lpm_key) -					\
+	    sizeof(((struct path_lpm_key *)0)->prefixlen))
+#define PATH_LPM_PREFIXLEN(_b)						\
+	(MIN(PATH_LPM_KEYLEN,						\
+	    ((_b) + sizeof(((struct path_lpm_key *)0)->meta))) * 8)
+
 extern void bpf_preempt_disable(void) __ksym __weak;
 extern void bpf_preempt_enable(void) __ksym __weak;
 
@@ -60,16 +70,16 @@ struct eval {
 	/* fields is the bitmask of things that have been filled */
 	__u64			 fields;		/* QUARK_RF_* bitmask */
 	__u64			 poison_tag;		/* QUARK_RF_POISON */
-	__u32			 pid;			/* QUARK_RF_PROCESS_PID */
-	__u32			 ppid;			/* QUARK_RF_PROCESS_PPID */
-	__u32			 uid;			/* QUARK_RF_PROCESS_UID */
-	__u32			 gid;			/* QUARK_RF_PROCESS_GID */
-	__u32			 sid;			/* QUARK_RF_PROCESS_SID */
+	__u32			 pid;			/* QUARK_RF_PID */
+	__u32			 ppid;			/* QUARK_RF_PPID */
+	__u32			 uid;			/* QUARK_RF_UID */
+	__u32			 gid;			/* QUARK_RF_GID */
+	__u32			 sid;			/* QUARK_RF_SID */
 	__u32			 pad0;
-	struct path_lpm_key	*process_filename;	/* QUARK_RF_PROCESS_FILENAME */
-	struct path_lpm_key	*file_path;		/* QUARK_RF_PROCESS_FILE_PATH */
+	struct path_lpm_key	*exe;			/* QUARK_RF_EXE */
+	struct path_lpm_key	*filepath;		/* QUARK_RF_FILEPATH */
 	struct nova_rule	*match;			/* result of eval_run */
-	char			 comm[16];		/* QUARK_RF_PROCESS_COMM */
+	char			 comm[16];		/* QUARK_RF_COMM */
 };
 
 static void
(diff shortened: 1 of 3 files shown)


nova.bpf.c
```diff
@@ -60,16 +70,16 @@ struct eval {
...
-	__u32			 sid;			/* QUARK_RF_PROCESS_SID */
+	__u32			 pid;			/* QUARK_RF_PID */
+	__u32			 ppid;			/* QUARK_RF_PPID */
+	__u32			 uid;			/* QUARK_RF_UID */
+	__u32			 gid;			/* QUARK_RF_GID */
+	__u32			 sid;			/* QUARK_RF_SID */
 	__u32			 pad0;
-	struct path_lpm_key	*process_filename;	/* QUARK_RF_PROCESS_FILENAME */
-	struct path_lpm_key	*file_path;		/* QUARK_RF_PROCESS_FILE_PATH */
+	struct path_lpm_key	*exe;			/* QUARK_RF_FILENAME */
```
nicholasberlin:
```suggestion
	struct path_lpm_key	*exe;			/* QUARK_RF_EXE */
```
right?

haesbaert: 100% :heart:

nicholasberlin (approved):

nicholasberlin:
Here's what my bot reviewer thinks:
# PR Review: #316 — Improve prefixlen calculation and fix executable matching

**Author:** haesbaert
**Branch:** mulu -> main
**Commits:** 1 | **Files changed:** 3 | **+49 -32**
**URL:** https://github.com/elastic/quark/pull/316

## Summary

Two changes:
1. Real bugfix: switch the LSM hook from `lsm/task_alloc` to `lsm/bprm_check_security` so `process.exe` rules match the binary actually being executed (`bprm->file`) instead of the parent task's existing `mm->exe_file`.
2. Refactor: introduce `PATH_LPM_PREFIXLEN(len)` macro and pass the actual `bpf_d_path` length (which includes NUL) instead of always passing the maximum key length. Also rename `process_filename`/`file_path` -> `exe`/`filepath` to match the userspace constant rename.

The LSM hook switch is correct and meaningful; the prefixlen change is a clean-up rather than a behavior fix. PR is small and focused.

## Findings

### Significant

**1. `lsm/bprm_check_security` semantic change in `eval_init_task` worth documenting.**
Switching from `task_alloc` (fork-time) to `bprm_check_security` (exec-time) means the task fields populated via `eval_init_task` now describe the *calling* task at exec time, not the new image:
- `comm` is still the parent shell's comm; it'll be replaced by the binary name only after this hook returns success.
- `uid`/`gid` are the calling task's (pre-exec) credentials, which differ from what they will be for setuid/setgid binaries.
- `pid`/`ppid` are stable across exec, no change.

That's a fine and arguably the right model for "match on the executable being launched", but rules written against the old `task_alloc` hook (matching `comm` against the new binary's name, for instance) will silently change meaning. The PR description briefly mentions "This will likely change in the future, since we will likely want to match on both" — worth at least a code comment near `eval_init_task` calling out which task context is captured at the bprm hook.

**2. `PATH_LPM_PREFIXLEN(len)` is functionally equivalent to the previous `PATH_LPM_KEYLEN * 8`, not a fix.**
The PR description frames this as an improvement, but for both exact and wildcard rules the LPM trie returns the same set of matches whether the lookup prefixlen is the path-length-in-bits or the maximum (2016). With max prefixlen, every stored entry was eligible by length and matched/didn't match purely on its stored bits; with the new tighter prefixlen, entries whose stored prefix exceeds the lookup length are excluded — but that excluded set was already excluded by the bit comparison itself (because the lookup buffer had a NUL terminator past the path, which can't equal a stored exact-match's path bytes).

I checked a few cases and couldn't construct one where the two approaches differ. So this is a tidiness change, which is fine — but the PR title and description suggest a behavior fix, which is actually finding (1) only. Consider clarifying the commit message: the bugfix is the LSM hook switch; the prefixlen rework is cleanup.

### Minor

- **`nova.bpf.c:79`**: comment `/* QUARK_RF_FILENAME */` next to `struct path_lpm_key *exe` is stale — the rename made the constant `QUARK_RF_EXE`. (The corresponding `filepath` line on the next line is correctly updated.)
- **`nova.bpf.c:21`**: `#define MAX(...)` is added but unused. Drop it (or move both `MAX`/`MIN` into a shared helpers header next time something needs them).
- **`nova.bpf.c:24-29`**: the `sizeof(((struct path_lpm_key *)0)->prefixlen)` and `->meta` idioms are correct and avoid hard-coded `4`/`2`, but they're noisy. Since `path_lpm_key` is the only consumer, a comment showing the algebra (`KEYLEN = sizeof(meta) + sizeof(path) = 2 + 250 = 252` bytes) would help future readers.
- **`nova.bpf.c:226`**: `eval.exe->prefixlen = PATH_LPM_PREFIXLEN(len)` is set on the scratch buffer pointer; the corresponding `prefixlen` writes inside `eval_loop` for `exe` and `filepath` were dropped in this PR (lines 134, 142 area). Make sure no callers leave `exe->prefixlen` unset before the lookup — for `exe` this is set in the hook before `eval_run` is called, so it's fine, but `filepath` currently has no producer setting `prefixlen` (or `eval->fields |= QUARK_RF_FILEPATH`), so the `if (eval->fields & QUARK_RF_FILEPATH ...)` branch in `eval_loop` is dead in this PR. That's expected since filepath isn't wired up yet, but worth either a `TODO` comment near the `filepath` lookup branch, or a note that the `prefixlen` write moved out of `eval_loop` and into the producer site by design.
- **`quark-test.c:2179-2189`**: new ruleset entries `/usr/bin/b*` etc. exercise wildcards alongside exact matches. Fine for the placeholder. The `/somethingreallyreallylong` entry is a nice probe of behavior for never-existent paths.
- The `noexe:` label naming convention works but `goto out` or `goto eval` would read more naturally — `noexe` reads like "we have no exe" but the label is reached on both success and failure paths. Subjective.

## Verdict

**APPROVE** — the LSM hook switch is the real fix and is correct; the prefixlen macro work is a clean-up that doesn't change behavior. Address the stale comment on line 79 and ideally drop unused `MAX`. Findings 1–2 are clarifications, not blockers.

Issue #317, opened 2026-04-28
Split quark_ruleset_match into something like quark_ruleset_poison_cache()


haesbaert:
On quark_queue_open() we do this:
```
	/*
	 * Pass rules on all processes, so we can have match on poison rules
	 */
	if (qq->ruleset != NULL) {
		struct quark_event	 qev;
		struct quark_process	*qp;

		bzero(&qev, sizeof(qev));

		RB_FOREACH(qp, process_by_pid, &qq->process_by_pid) {
			qev.process = qp;
			quark_ruleset_match(qq->ruleset, &qev);
		}
	}
```

We do this really just to poison all things, but we also execute basically all targets, the problem is that it screws up with counters, for example, imagine this test case
```
	if (asprintf(&text_ruleset,
	    "pass on process.pid %d file.path /tmp/quark-test-suff*\n"
	    "drop on process.pid %d\n",
	    getpid(), getpid()) == -1)
		err(1, "asprintf");
```

At boot, you would never expect the rule count for the drop rule to be > 0, but it is, because we evaluated everything under a fake process. This is confusing AF when writing tests.

PR #318, opened 2026-04-28
Expand the ruleset to accept infix path matching

haesbaert:
Extend the grammar to allow matching on paths like /home/*/foo, or /h**e/. This
is limited to a single *, this allows us to implement the matching quite fast
without backtracking and/or a regex machine.

This limitation will also make implementing this in bpf much easier.

Had to bloat quark_rule_field{} a bit with a quark_wild{} since we have to
describe way more things now. The structure grows considerably, expressing a
path took 16 bytes before, now it takes 32 :(. And it grows for every field
case, so very meh.

Next diffs will be adding this to QQ_NOVA.

Issue #316

+++

Only bump rule counters for poison rules on quark_queue_open()

On boot at quark_queue_open() we evaluate all rules against our cached process
for the sole purpose of poisoning processes.

While this 'works', it screwed up the rule counters and made writing tests for
it more painful. So make sure we reset those counters.

Issue #316
Issue #317

Issue #315

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -276,7 +276,7 @@ steps:
 
   - label: "quark-test on rhel 8.5 (no file)"
     key: test_rhel_8_5
-    command: "./.buildkite/runtest_distro.sh rhel 8.5 -x t_file_bypass -x t_file -x t_memfd -x t_shm_open -x t_rule_path"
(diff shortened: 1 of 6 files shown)


haesbaert:
So the bot spotted a case I missed when evaluating the poison rules on boot, so I reverted that and now I just reset the counters.

nicholasberlin (approved):

PR #322, opened 2026-04-29
Fall back to renamed VFS probe symbols 

nicholasberlin:
Newer kernels rename a few VFS helpers we attach to:
do_unlinkat → filename_unlinkat,
do_renameat2 → filename_renameat2.
do_filp_open → do_file_open,

The prototypes are unchanged, so the BPF program bodies are identical.

Add twin SEC entries in Probe.bpf.c that share the existing
__enter/__exit helpers, and BTF-detect each old name in
bpf_queue_open1() to autoload the right variant.

Relevant Linux commits:
https://github.com/torvalds/linux/commit/e50aae1d39ac37a95f453a699456b73dd07e3913
https://github.com/torvalds/linux/commit/e6d50234ccb9ff54addd579032a146aef52e7541
https://github.com/torvalds/linux/commit/541003b576c3e3c328314398a6df76eb3cebf847

Also, enabled 26.04 testing.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -494,13 +494,13 @@ steps:
       machineType: n2-standard-2
       enableNestedVirtualization: true
 
-#  - label: "quark-test on ubuntu 26.04"
-#    key: test_ubuntu_26_04
-#    command: "./.buildkite/runtest_distro.sh ubuntu 26.04"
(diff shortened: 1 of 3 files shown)


haesbaert:
love it

haesbaert (approved):

PR #329, opened 2026-05-06
Executing in-memory payloads via memfd_create + execve("/proc/self/fd/<N>") produces incorrect values

nicholasberlin:
Two commits, the first addresses the title of the PR. The second addresses verifier issues probably stemming from my dev box using clang-14.

### DIFF
--- a/elastic-ebpf/GPL/Events/PathResolver.h
+++ b/elastic-ebpf/GPL/Events/PathResolver.h
@@ -66,13 +66,33 @@ ebpf_resolve_path_to_string(char *buf, struct path *path, const struct task_stru
     // All struct vfsmount's are stored in a struct mount. We need fields in
     // the struct mount to continue the dentry walk when we hit the root of a
     // mounted filesystem.
-    struct mount *mnt          = container_of(curr_vfsmount, struct mount, mnt);
-    struct dentry *curr_dentry = BPF_CORE_READ(path, dentry);
+    struct mount *mnt                    = container_of(curr_vfsmount, struct mount, mnt);
+    struct dentry *curr_dentry           = BPF_CORE_READ(path, dentry);
+    struct dentry *leaf_parent           = BPF_CORE_READ(curr_dentry, d_parent);
+    const struct dentry_operations *d_op = BPF_CORE_READ(curr_dentry, d_op);
     struct dentry **dentry_arr;
 
     // Ensure we make buf an empty string early up here so if we exit with any
     // sort of error, we won't leave garbage in it if it's uninitialized
     buf[0] = '\0';
+
+    // Pseudo / anonymous dentries (memfd, anon_inode, pipefs, sockfs, ...)
+    // are self-parented and live on kernel-internal mounts that are never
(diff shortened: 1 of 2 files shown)


nicholasberlin:
Looks like I need to revert the verifier changes and use a newer clang on my VM.

elastic-ebpf/GPL/Events/PathResolver.h
```diff
@@ -74,6 +74,29 @@ ebpf_resolve_path_to_string(char *buf, struct path *path, const struct task_stru
...
 
+    // Pseudo / anonymous dentries (memfd, anon_inode, pipefs, sockfs, ...)
+    // are self-parented and live on kernel-internal mounts that are never
+    // grafted onto the task's mount tree. Walking the parent chain spins
+    // until truncation and produces garbage. The kernel's d_path() handles
+    // these via dentry->d_op->d_dname; mirror that here by emitting d_name
+    // directly. For memfd execs this yields "memfd:<name>", matching what
+    // d_path() emits (minus the " (deleted)" suffix simple_dname appends).
+    {
+        struct dentry *leaf_parent           = BPF_CORE_READ(curr_dentry, d_parent);
```
haesbaert: no need for a new block, the rest of the code doesn't do this so I'd rather not do it.

quark-test.c
```diff
@@ -525,6 +525,76 @@ fork_n(int n)
...
+
+		if (snprintf(fdpath, sizeof(fdpath), "/proc/self/fd/%d", fd)
+		    >= (int)sizeof(fdpath))
+			errx(1, "snprintf fdpath");
+
+		execv(fdpath, argv);
+		err(1, "execv memfd");
+	}
+	/* NOTREACHED */
+	return (0);
```
haesbaert: This can be simplified by using loadfile_nostat + qwrite(), I'll push a diff

haesbaert (changes requested):
Thanks for fixing it!

Test looks good, but it can be simplified a bit, I'm pushing a change.

I'm not ok with the second diff to make it build on old clang, since it breaks other things, I'd take that back and update your clang if possible, if not we should find a way to fix for all cases, or just stop supporting multiple clang and stick to one?

haesbaert:
> Looks like I need to revert the verifier changes and use a newer clang on my VM.

I think so :/. Maybe we should pinpoint one specific version and support just that? 

At any rate, we will have to update zig in the $OTHER_PRODUCT.

haesbaert:
> > Looks like I need to revert the verifier changes and use a newer clang on my VM.
> 
> I think so :/. Maybe we should pinpoint one specific version and support just that?
> 
> At any rate, we will have to update zig in the $OTHER_PRODUCT.

Btw, you can get by in the dev box with make CLANG=clang-18, assuming you have access to a clang-18

elastic-ebpf/GPL/Events/PathResolver.h
nicholasberlin: forced pushed a change

haesbaert (approved):

PR #335, opened 2026-05-18
Populate the ringbuffer in QQ_NOVA

haesbaert:
```
Date:   Mon May 18 13:31:23 2026 +0200

    Populate the ringbuffer in QQ_NOVA

    This diff starts putting things into the ringbuffer, which means also adding the
    code for retrieving "most" of the task fields so far, see task_to_nova().

    There are still missing fields, like cwd and env, these will come later.

    The API for outputing things in the ringbuffer is called "output", it mimicks
    the same underlying helpers:
        ouput_reserve -> bpf_ringbuf_reserve_dynptr
        output_discard -> bpf_ringbuf_discard_dynptr
        output_submit -> bpf_ringbuf_submit_dynptr

    The API does a bit more than just calling the helper, and we also do it so we
    can support kernels that don't have dynptr ([5.17:5.19[). In the future I can
    add code below this API to handle the case without dynptr.

    Why is dynptr cool?
    The old bpf_ringbuf_reserve() needs a compile time constant for reserve(),
    meaning we can't reserve some length bytes stored in a variable, it needs to be
    a compile time constant. That's why the old probes have to copy everything in a
    temporary buffer, then do bpf_ringbuf_output(), because that one accepts a
    runtime length.

    Dynptr solves exactly that, the verifier knows how to track then length so we
    can use it in runtime, which means saving _full_ copy of the event for
    every event.

    We do variable length data, like paths and long strings, with a struct nova_vl,
    which can be written with output_vl_write(). nova_vl is stashed inside the
    actual event and contains offset and length so userland can call vltoc() (see
    nova_queue.c). This fixes a long standing grip I had with the old probes where
    you had to loop to find fields.

    struct eval{} now has the task fields in struct nova_task nt, we assemble nt
    once, which later is copied to the output buffer. This kinda sucks, and I tried
    really hard to do evaluation inside the ringbuffer itself, but it's not feasible
    for a myriad of reasons (trust me bro), the main one being we would have to
    reserve the maximum possible size of an event.

```

### DIFF
--- a/nova.bpf.c
+++ b/nova.bpf.c
@@ -7,12 +7,16 @@
 #include <bpf/bpf_tracing.h>
 #include <bpf/bpf_core_read.h>
 
+/* XXX Remove me once we update our ancient vmlinux.h XXX */
+struct bpf_dynptr {
+	__u64 __opaque[2];
+} __attribute__((aligned(8)));
+
 #include "nova.h"
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic error "-Wpadded"
 
-/* XXX check if there isn't something like this on the headers XXX */
 #define LOOP_CONTINUE	0
 #define LOOP_STOP	1
 
@@ -26,6 +30,14 @@
 
 extern void bpf_preempt_disable(void) __ksym __weak;
 extern void bpf_preempt_enable(void) __ksym __weak;
+
+struct output {
+	__u32			head_len;
+	__u32			total_len;
+	__u32			var_off;
+	__u32			pad;
+	struct bpf_dynptr	dptr;
+};
 
 struct eval_path {
 	char			 full[PATH_MAX];
@@ -42,17 +54,11 @@ struct eval_path {
  */
 struct eval {
 	/* fields is the bitmask of things that have been filled */
-	__u64			 fields;		/* QUARK_RF_* bitmask */
-	__u64			 poison_tag;		/* QUARK_RF_POISON */
-	__u32			 pid;			/* QUARK_RF_PID */
-	__u32			 ppid;			/* QUARK_RF_PPID */
-	__u32			 uid;			/* QUARK_RF_UID */
-	__u32			 gid;			/* QUARK_RF_GID */
-	__u32			 sid;			/* QUARK_RF_SID */
-	__u32			 pad0;
-	struct eval_path	 exe;			/* QUARK_RF_EXE */
-	struct eval_path	 filepath;		/* QUARK_RF_FILEPATH */
-	char			 comm[16];		/* QUARK_RF_COMM */
+	__u64			fields;		/* QUARK_RF_* bitmask */
+	__u64			poison_tag;	/* QUARK_RF_POISON */
+	struct nova_task	nt;
+	struct eval_path	exe;		/* QUARK_RF_EXE */
+	struct eval_path	filepath;	/* QUARK_RF_FILEPATH */
 };
 
 struct {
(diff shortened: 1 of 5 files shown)


nova.bpf.c
```diff
@@ -286,25 +449,42 @@ bprm_check1(struct linux_binprm *bprm, int ret)
...
 noexe:
 	r = eval_run(eval);
-	r = 0; /* XXX r ignored for now */
 
-	return (r);
+	if (r != QUARK_RA_PASS)
+		goto done;
+
+	exec = output_reserve(&o, sizeof(*exec),
+	    sizeof(*exec) + exe->full_len);
```
nicholasberlin, line 466: `exe` can be uninitialized here if line 450 hits, right?

nova.bpf.c
```diff
@@ -253,31 +363,84 @@ eval_init(void)
...
+static __always_inline void *
+output_reserve(struct output *o, __u32 head_len, __u32 total_len)
+{
+	struct nova_event	*ev;
+
+	o->head_len = head_len;
+	o->total_len = total_len;
+	o->var_off = o->head_len;
+
+	if (bpf_ringbuf_reserve_dynptr(&output_ring,
```
nicholasberlin, line 379: The bot thought this could be a problem because failures would leave `o->dptr` uninitialized. But, it looks like all the failure paths set it to null. https://github.com/torvalds/linux/blob/master/kernel/bpf/ringbuf.c#L670-L699

nicholasberlin (approved):
The bot pointed out that there's no rule support for RF_COMM despite comm being populated now. 

> Nova still accepts QUARK_RF_COMM rules, but the BPF evaluator can no longer satisfy them. nova_rule_from_quark() copies field->comm and sets nr->fields |= QUARK_RF_COMM, while eval_init_task() now only marks PID/PPID/UID/GID as present, and eval_loop() has no QUARK_RF_COMM comparison against eval->nt.comm. Any Nova ruleset containing a comm predicate will never match even though task_to_nova() now fills nt.comm. Either implement the comm comparison and mark the field as present, or reject/strip unsupported QUARK_RF_COMM rules for Nova until that support lands.

nova.bpf.c
haesbaert: not really because:
```
static void
eval_path_init(struct eval_path *ep)
{
	ep->full_len = 0;
	ep->full[0] = 0;
	ep->pre.prefixlen = PATH_LPM_KEY_BITLEN;
	ep->post.prefixlen = PATH_LPM_KEY_BITLEN;
}
```
we always do eval_path_init() via eval_init()

nova.bpf.c
haesbaert: oh yeah the bot is wrong, the semantics is to call discard even if the initial reserve fails, the verifier will reject it.

haesbaert:
> The bot pointed out that there's no rule support for RF_COMM despite comm being populated now.
> 
> > Nova still accepts QUARK_RF_COMM rules, but the BPF evaluator can no longer satisfy them. nova_rule_from_quark() copies field->comm and sets nr->fields |= QUARK_RF_COMM, while eval_init_task() now only marks PID/PPID/UID/GID as present, and eval_loop() has no QUARK_RF_COMM comparison against eval->nt.comm. Any Nova ruleset containing a comm predicate will never match even though task_to_nova() now fills nt.comm. Either implement the comm comparison and mark the field as present, or reject/strip unsupported QUARK_RF_COMM rules for Nova until that support lands.

aye, there's more stuff missing, I wouldn't mind it much now, once stuff is in place and I can write proper tests I'll catch all those.

haesbaert:
Thanks for the review!

Today while walking back from kindergarden I had another idea.
It might make the code less confusing, if we just have a temporary output buffer, like the old probes, then we put the event that we want to output there, THEN we do the eval_run() with a pointer to the output buffer, so we essentially do an eval on what we're about to output + some metadata that will be needed.

This would have the same number of copies:
today: copy into eval, copy into ring buffer
with what I thought: copy into the output buffer, copy output buffer into ring buffer.

The fact that we need to have a temporary buffer since we need to know the paths len, kinda defeats the purpose of using dynptr, so we might as well just not use it, it might also make the code easier to read.

PR #336, opened 2026-05-21
Give probes running in softirq context their own output buffer.

haesbaert:
Turns out bpf probes do not disable bottom half/softirq interrupts.
All probes use a per cpu output event buffer for staging the data to be put in
the ringbuffer.

Since softirq probes (cgroup_skb/egress, cgroup_skb/ingress) can run on top of a
kprobe/fentry probe (all the other ones), they end up racing the event buffer.

This diff makes softirq probes use their own output buffer event_buffer_map_softirq.

### DIFF
--- a/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
@@ -296,6 +296,9 @@ static int skb_peel_nexthdr(struct __sk_buff *skb, u8 wanted)
 }
 #endif
 
+/*
+ * See cgroub_skb/egress, pretty please.
+ */
 static int skb_in_or_egress(struct __sk_buff *skb, int ingress)
 {
     struct udphdr udp;
@@ -344,10 +347,10 @@ static int skb_in_or_egress(struct __sk_buff *skb, int ingress)
         goto ignore;
 
     /*
-     * Needed for kernels prior to f79efcb0075a20633cbf9b47759f2c0d538f78d8
+     * Needed for kernels prior to cd17d38f8b28f808c368121041c0a4fa91757e0d
      * bpf: Permits pointers on stack for helper calls
      */
-    sk_addr = bpf_map_lookup_elem(&scratch64, &zero);
+    sk_addr = bpf_map_lookup_elem(&scratch64_softirq, &zero);
     if (sk_addr == NULL)
         goto ignore;
     *sk_addr = (u64)sk;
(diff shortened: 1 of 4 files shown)


haesbaert:
This is ready for some bashing, I just want to write a program that exposes the crash, I have one that exposes the bug, but want to confirm.

elastic-ebpf/GPL/Events/Network/Probe.bpf.c
nicholasberlin: think we should give the same treatment to `scratch64`? I see it being accessed in the other cgroup probes.

haesbaert: absolutely! zefixed

nicholasberlin (approved):

haesbaert:
For reference, this is what I used to trigger the race more easily:
```
commit 904bca14f680c773d9386f414a747ecf57f891fe
Date:   Thu May 21 20:54:19 2026 +0200

    bang

diff --git elastic-ebpf/GPL/Events/Varlen.h elastic-ebpf/GPL/Events/Varlen.h
index f30de8b..f3861aa 100644
--- elastic-ebpf/GPL/Events/Varlen.h
+++ elastic-ebpf/GPL/Events/Varlen.h
@@ -62,8 +62,12 @@ struct {
 
 static void *get_event_buffer()
 {
+	void *p;
     int key = bpf_get_smp_processor_id();
-    return bpf_map_lookup_elem(&event_buffer_map, &key);
+    p = bpf_map_lookup_elem(&event_buffer_map, &key);
+    if (p)
+	    __builtin_memset(p, 0xff, 1024);
+    return p;
 }
 
 struct ebpf_varlen_field *ebpf_vl_field__add(struct ebpf_varlen_fields_start *fields,

```

```
make && sudo ./quark-mon -vb -GFSs -N >/dev/null
```

And then a bunch of cpus doing things like hostname lookups, /bin/whoami, /bin/echo foo > /tmp/mulumulu

With that, as long as DNS probes were loaded we could corrupt the event:
```
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
quark-mon: ebpf_events_to_raw:817: unhandled type 18446744073709551615
```

PR #339, opened 2026-05-22
Use nested context for dns probes

haesbaert:
Things are worse than f6284da030a07912131b9e17ac7c1efc48f80e3b

These probes can actually trigger from process context as well as softirq
context.

The most normal path for skb_egress is to trigger from process context:
write(2) -> sock_write().... -> udp_output() -> ip_output()....
or whatever the hell linux calls it.

But skb_egress can also trigger from an actual softirq, imagine sending a
TCP_ACK for an incoming packet that is being processed at softirq.
Important: not only skb_egress can race against skb_ingress, skb_egress can race
against skb_egress as well! The bottom half is not disabled in the output path.

The skb_ingress path seems to be similar, it can trigger both from softirq and
process context. The softirq is the normal case, think of a packet hitting the
card and then dispatching a softirq. The process context can happen when a
write() to localhost short-circuits two processes.

Instead of trying to figure out all the possible combinations, rewrite the DNS
events to use nested contexts, even though we only need 2, use 3, because well,
because yes. This means any combination with anything should work well.

### DIFF
--- a/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
@@ -29,6 +29,50 @@ struct {
     __uint(map_flags, BPF_F_NO_PREALLOC);
 } sk_to_tgid SEC(".maps");
 
+struct dns_ctx {
+    /*
+     * which nest index we are, only meaningful on the first one,
+     * saves having a map just to store an int
+     */
+    u32 nest;
+    u64 scratch64;
+    struct ebpf_dns_event event; /* must be last */
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(key_size, sizeof(u32));
+    __uint(value_size, sizeof(struct dns_ctx) + MAX_DNS_PACKET);
+    __uint(max_entries, 3);
+} dns_ctx_map SEC(".maps");
+
+struct dns_ctx *
+get_dns_ctx(void)
+{
+    struct dns_ctx *ctx, *zero_ctx;
+
+    zero_ctx = bpf_map_lookup_elem(&dns_ctx_map, &(int){0});
+    if (zero_ctx == NULL)
+        return (NULL);
+    ctx = bpf_map_lookup_elem(&dns_ctx_map, &zero_ctx->nest);
+    if (ctx == NULL)
+        return (NULL);
+    zero_ctx->nest++;
+
+    return (ctx);
+}
+
+void
+put_dns_ctx(void)
+{
+	struct dns_ctx *zero_ctx;
+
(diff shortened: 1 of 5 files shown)


elastic-ebpf/GPL/Events/Network/Probe.bpf.c
```diff
@@ -438,19 +477,22 @@ int skb_egress(struct __sk_buff *skb)
...
 
-    return r;
+    return 1;
 }
 
 static int sk_maybe_save_tgid(struct bpf_sock *sk)
 {
     u32 tgid, zero = 0;
-    u64 *sk_addr;
+    struct dns_ctx *ctx;
```
nicholasberlin, line 495: don't need zero on the line above anymore.

nicholasberlin (approved):
LGTM. The bot agrees:

# PR Review: #339 - Use nested context for dns probes

**Author:** haesbaert
**Branch:** nest -> main
**Commits:** 1 | **Files changed:** 5 | **+86 -85**
**URL:** https://github.com/elastic/quark/pull/339
**Reviewed commit:** f9d7da14cab65c23a14d560c4eaa4b018881d5d4
**Reviewed by:** GPT-5.5

## Summary

This PR moves DNS probe scratch/event storage into a small nested per-CPU context so cgroup DNS paths can tolerate process-context and softirq interleaving, including egress nesting. The BPF event layout change is matched by the userspace decoder, and the get/put paths for the nested context look balanced.

## Findings

No confirmed correctness, security, performance, or maintainability issues found.

## Verdict

APPROVE - no code changes requested. Wait for CI before merging.

elastic-ebpf/GPL/Events/Network/Probe.bpf.c
```diff
@@ -29,6 +29,50 @@ struct {
...
+struct dns_ctx *
+get_dns_ctx(void)
+{
+    struct dns_ctx *ctx, *zero_ctx;
+
+    zero_ctx = bpf_map_lookup_elem(&dns_ctx_map, &(int){0});
+    if (zero_ctx == NULL)
+        return (NULL);
+    ctx = bpf_map_lookup_elem(&dns_ctx_map, &zero_ctx->nest);
+    if (ctx == NULL)
```
nicholasberlin, line 58: it we are interrrupted here, won't the same ctx be returned?

haesbaert: it will, but that interrupt will do all its thing then return, and we will be on the same nest level we were before, does that make sense?

nicholasberlin: yes, i imagined them both proceeding into code at the same time, instead of the interrupt running to completion. Thanks.

LGTM

PR #341, opened 2026-05-29
Emit fork from a trace probe on nova

haesbaert:
Two relatively easy commits

commit 1a377c7ecaefe7bf187cb7469b052a7b4684b53b
Date:   Fri May 29 08:48:09 2026 +0200

    Nova: preempt_{disable,enable} -> probe_{prologue,epilogue}

    Not just a rename, preparation for having more things, like counters and
    whatnot.

commit fe31a2785ba24e0fe5e0a592c1555f72374577c1
Date:   Fri May 29 08:42:39 2026 +0200

    Emit fork from a trace probe on nova

    task_alloc()@lsm happens pre pid allocation so we're basically sending a parent
    event, so move it to a trace probe.

    There's still things missing, especially cwd, which we might need to stash
    somewhere and retrieve as we can't bpf_d_path() there.

    Issue #315

### DIFF
--- a/nova.bpf.c
+++ b/nova.bpf.c
@@ -529,6 +529,49 @@ BPF_PROG(bprm_check, struct linux_binprm *bprm, int ret)
 }
 
 /*
+ * TODO stop evaluating things twice, store the result from the lsm probe
+ * somewhere and retrieve here.
+ */
+static int
+sched_process_fork1(struct task_struct *parent, struct task_struct *child)
+{
+	struct eval		*eval;
+	int			 r;
+
+	if ((eval = eval_init()) == NULL)
+		return (0);
+	eval_init_task(eval, child);
+
+	r = eval_run(eval);
+
+	if (r != QUARK_RA_PASS)
+		goto done;
+
+	/*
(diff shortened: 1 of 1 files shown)


nova.bpf.c
```diff
@@ -521,13 +533,56 @@ BPF_PROG(bprm_check, struct linux_binprm *bprm, int ret)
...
+	int			 r;
+
+	if ((eval = eval_init()) == NULL)
+		return (0);
+	eval_init_task(eval, child);
+
+	r = eval_run(eval);
+
+	if (r != QUARK_RA_PASS)
+		goto done;
```
nicholasberlin, line 560: return (0); for consistency?

nicholasberlin (approved):

ricardoamaro (approved):

nova.bpf.c
haesbaert: I'll do a cleanup and fix this, we don't even need the label.
But as we discussed, a lot has to change since the lsm probes are too early

haesbaert:
Issue #315

Issue #330, opened 2026-05-13
Consider having nightly builds


haesbaert:
Would be nice to have nightly builds here somehow, when customers have exoteric kernels, we end up asking support to build and test so we can see why the probes are not loading.
If we had nightly builds in releases, we could more easily point it out.
As long as it wasn't some faulty github thingy and we had a clean single nightly release that doesn't stay in the way of everything else, it might be worth it.

I know zilch about github trickery.

haesbaert:
The important bits are the binaries, so people can test it. We can also include the header files and an archive, but actual software using quark is expected to build their own libquark.a.

ricardoamaro:
@gogochan @dwhyrock for viz

gogochan:
For now, let's proceed with a new buildkite pipeline and publish nightly build artifact to github.

PR #345, opened 2026-06-16
Add a target for quark-bin.tar.gz

haesbaert:
Related to issue #330

Trivial target for having all our binaries in a tarball so we can have a nightly
build in github.


haesbaert:
cc @gogochan

gogochan (approved):

PR #346, opened 2026-06-16
Added a github workflow for building and publishing quark-bin.tar.gz nightly

gogochan:
This PR adds a github workflow for building and publishing quark-bin.tar.gz nightly.

This workflow can only be triggered by users with admin role for now.

Following the same rules in Makefile to build CentOS and Ubuntu containers and use them to build quark. 

I'm not 100% sure, if I have to build this from CentOS 7. I see that quark-builder is using Ubuntu 26.04

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -18,11 +18,11 @@ jobs:
 
     steps:
       - name: Check actor permissions
-        if: github.event_name == 'workflow_dispatch'
+        if: github.event_name == 'workflow_dispatch' && inputs.publish == true
         run: |
           ROLE=$(gh api repos/${{ github.repository }}/collaborators/${{ github.actor }}/permission --jq '.role_name')
           if [[ "$ROLE" != "admin" ]]; then
-            echo "Only repository admins can trigger this workflow manually (got: $ROLE)"
+            echo "Only repository admins can publish a nightly release (got: $ROLE)"
             exit 1
           fi
         env:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -13,14 +13,77 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
     steps:
-      - name: Stub
-        run: echo "nightly build stub"
+      - name: Check actor permissions
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          ROLE=$(gh api repos/${{ github.repository }}/collaborators/${{ github.actor }}/permission --jq '.role_name')
+          if [[ "$ROLE" != "admin" ]]; then
+            echo "Only repository admins can trigger this workflow manually (got: $ROLE)"
+            exit 1
+          fi
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - name: Build Ubuntu Docker image
+        run: make docker-image
+
+      - name: Build CentOS 7 Docker image
+        run: make centos7-image
+
+      - name: Build BPF probes in Ubuntu container
+        run: |
+          docker run \
+            -v "$(pwd):$(pwd)" \
+            -w "$(pwd)" \
+            -u "$(id -u):$(id -g)" \
+            quark-builder \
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Build quark-bin.tar.gz in CentOS 7 container
+        run: |
+          docker run \
+            -v "$(pwd):$(pwd)" \
+            -w "$(pwd)" \
+            -u "$(id -u):$(id -g)" \
+            -e CENTOS7=y \
+            centos7-quark-builder \
+            sh -c "make -j1 NO_GO=y -C $(pwd) quark-bin.tar.gz"
```
haesbaert: Hmm not a big fan of having all this logic in a CI pipeline, also I'm not convinced we need it.

The build should really just do `make quark-bin.tar.gz`, nothing more.
If it doesn't work, leme know and I'll fix it, but I did make it so that you should be able to do `make quark-bin.tar.gz` and it automatically builds only what it needs to fulfill the target.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Check actor permissions
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          ROLE=$(gh api repos/${{ github.repository }}/collaborators/${{ github.actor }}/permission --jq '.role_name')
+          if [[ "$ROLE" != "admin" ]]; then
+            echo "Only repository admins can trigger this workflow manually (got: $ROLE)"
+            exit 1
+          fi
+        env:
+          GH_TOKEN: ${{ github.token }}
```
haesbaert, line 32: I don't understand CI enough, but I trust you, just noting I can't give a great feedback in this bit.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Upload nightly release
+        uses: softprops/action-gh-release@v3
+        with:
+          tag_name: nightly
+          name: Nightly Build
+          prerelease: true
+          make_latest: false
+          body: "Automated nightly build from commit ${{ github.sha }}"
+          files: quark-bin.tar.gz
```
haesbaert: If I understand correctly, this will push a tarball without the date in the name?
I think the date should be reflected somewhere, like quark-bin-$(date +"%m-%d-%y").tar.gz or something like it.
What do you think?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: quark-bin
+          path: quark-bin.tar.gz
+
+  release:
+    needs: build
+    if: github.event_name == 'schedule' || inputs.publish == true
```
haesbaert, line 56: This looks fishy, is there a case where the event_name is schedule and inputs.publish is false? I don't know what those are so just asking.

haesbaert (commented):
Thanks for working on this! It will help us a lot in debugging customer setups ❤️ 

I've added some comments, I'd be happy with just a `make quark-bin.tar.gz` as the build step. Adding docker run commands in the pipeline is too brittle as now I'd have to consider it everytime I change the makefile, so `make nightly` should be the interface for this PR.

.github/workflows/nightly.yml
gogochan: Is there any requirement that I have to build this from any particular OS? I see the docker builder uses Ubuntu 26.04, and there are some references to CentOS 7. It would simplify a lot, if I can just use Ubuntu 26.04.

.github/workflows/nightly.yml
gogochan: this gets auto populated by Github.

.github/workflows/nightly.yml
gogochan: Yes, let me change it. I was mainly modeling after neovim's nightly build, but I see other projects has timestamps in their artifacts and it would be easier to communicate with the customer.

.github/workflows/nightly.yml
gogochan: This only execute as a scheduled job, or if someone manually triggered it with publish option checked. The second path is guarded against non-admin user in line 20.

gogochan:
Let me create another PR with empty workflow to register the workflow to test first

gogochan:
I broke down the PR.
https://github.com/elastic/quark/pull/348 adds a dummy workflow.
This PR adds the actual implementation.

PR #347, opened 2026-06-17
Split out container_by_id tree from kubernetes structure

Tacklebox:
This moves the container tree into `quark_queue`. There is now a separate `linked` field for each of the `container_by_id` and `container_by_pod` trees.


haesbaert (approved):
Looks great!
Thanks for the PR

PR #351, opened 2026-06-19
Add commit checks to nightly build

gogochan:
This PR adds logic that checks the commit hash of the published nightly build to prevent unnecessary nightly builds.

Also added Force Build option to override it.
image

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -13,6 +13,10 @@ on:
         description: 'Enable verbose build output (V=1)'
         type: boolean
         default: false
+      force:
+        description: 'Force build even if commit hash has not changed'
+        type: boolean
+        default: false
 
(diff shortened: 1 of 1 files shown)


haesbaert:
This is ok, but I thought we would always only have one nightly build? I see no reason to have multiple ones, so whatever we build should replace the old one, not add to it, or I misread something?

haesbaert (approved):

gogochan:
> This is ok, but I thought we would always only have one nightly build? I see no reason to have multiple ones, so whatever we build should replace the old one, not add to it, or I misread something?

That is true, we always have only one nightly build. The current logic prevents you from building another one, if the hash matches. `Force build` lets you build even if the currently nightly has the same commit hash. I just added it for potential debugging scenario.

Issue #350, opened 2026-06-19
Hybrid cgroup sproc parsing is wrong


haesbaert:
hybrid cgroup /proc parsing is "wrong".

In a system with hybrid cgroups, when you do this: `GRUB_CMDLINE_LINUX_DEFAULT="systemd.unified_cgroup_hierarchy=0"`

We then parse /proc/$PID/cgroup for every process when we come up, and we get this:
```
->668
  COMM  comm=quark-mon
  CMDL  cmdline=[ ./quark-mon, -b ]
  PROC  ppid=667
  PROC  uid=0 gid=0 suid=0 sgid=0 euid=0 egid=0 pgid=668 sid=667
  PROC  cap_inheritable=0x0 cap_permitted=0x1ffffffffff cap_effective=0x1ffffffffff
  PROC  cap_bset=0x1ffffffffff cap_ambient=0x0
  PROC  time_boot=1781857315840000000 tty_major=136 tty_minor=1
  PROC  uts_inonum=4026531838 ipc_inonum=4026531839
  PROC  mnt_inonum=4026531841 net_inonum=4026531840
  PROC  entity_id=nAIAAAAAE7/JbboY, entry_leader_type=UNKNOWN entry_leader=0
  CWD   cwd=/home/haesbaert
  EXE   exe=/home/haesbaert/quark-mon
  ENV   env_len=2030
  CGRP  cgroup=/user.slice
12:cpuset:/
11:hugetlb:/
10:blkio:/
9:misc:/
8:perf_event:/
7:rdma:/
6:net_cls,net_prio:/
5:pids:/user.slice/user-1000.slice/session-1.scope
4:freezer:/
3:memory:/user.slice/user-1000.slice/session-1.scope
2:cpu,cpuacct:/
1:name=systemd:/user.slice/user-1000.slice/session-1.scope
0::/user.slice/user-1000.slice/session-1.scope  <---------------- BASICALLY WE WANT THIS ONLY
```

Subsequent created processes get the the v2 entry (the 0:: thing), since we read it from the kernfs_node:
```
->674 (FORK+EXEC+EXIT)
  COMM  comm=ls
  CMDL  cmdline=[ ls, --color=auto ]
  PROC  ppid=620
  PROC  uid=1000 gid=1000 suid=1000 sgid=1000 euid=1000 egid=1000 pgid=674 sid=620
  PROC  cap_inheritable=0x0 cap_permitted=0x0 cap_effective=0x0
[...]

PR #352, opened 2026-06-19
LXC container cgroup identification

Tacklebox:
This adds support for correlating events associated with specific LXC containers by their cgroup. It only works with the cgroup format from lxc 4.0+ but it should support legacy, hybrid, and unified cgroups.

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -2741,7 +2741,7 @@ sproc_cgroup(struct quark_process *qp, int dfd)
 {
 	int	 fd;
 	size_t	 len;
-	char	*cgroup, *p;
+	char	*cgroup, *tail, *p;
 
 	cgroup = NULL;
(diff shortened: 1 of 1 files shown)


haesbaert (approved):

quark.c
```diff
@@ -2758,21 +2758,19 @@ sproc_cgroup(struct quark_process *qp, int dfd)
 	/* if cgroup != NULL, len > 0 */
 	cgroup[len - 1] = 0;
 	/*
```
haesbaert: this will break the code below, it will remove the last line, so the your tail strstr will always fail if 0:: is the last line.
Just remove this line.

haesbaert (changes requested):

haesbaert (approved):

haesbaert:
related to #352

haesbaert:
fixes #350

PR #358, opened 2026-07-06
Adding initial TLS attach/detach mechanism

P1llus:
This PR adds the possibility to attach uprobes properly to both the systems OpenSSL shared library (libssl*.so) through symbols and to binaries which includes their own compiled static linked OpenSSL/BoringSSL but has their symbols stripped (through offsets).

This represents a draft/opinion on an implementation rather than a required feature request. A few important notes below:

**Notes:**
- It's purely an additive feature, behind a flag that is disabled by default. Any existing user of quark should be able to upgrade and recompile without doing anything and should not break their implementation. If you find anything that contradicts that then please let me know.
- This is purely the functionality itself, it does not include any logic on how to trigger it but rather just the API itself, that keeps the PR "smaller" but also allows for additions in either quark or others that rely on quark to decide their own logic.
- It is not the responsibility of this feature to locate correct offsets, protect against things like TOCTOU or parse any of the data collected. 
- It is the responsibility of the feature to fail safely (for example missing symbols or ensuring all required probes connect correctly), to provide the required arguments to cleanly connect the uprobes, and to properly initialize and destroy any references on startup/shutdown.
- It would in theory work on any protocol, so to not bloat quark's codebase the handling of the actual plaintext bytes is left to the consumer (another quark function or a queue consumer).
- I tried my best to reuse quark's existing functionality and design, including the current event queue, if anyone wants it differently then please let me know as I am open to any feedback or pointers to different ways to implement it.

I won't be including any direction on "how" to retrieve correct offsets, or when/how to trigger this in the PR description but would be happy to discuss it elsewhere.

# Summary

## Answered questions:
1. Q: Currently the default ringbuffer is 4MB, the "best" scenario would be to give this a separate ringbuffer so that it won't have the possibility to block other more important events.
A: Currently we can increase the existing ringbuffer and test first, a second ringbuffer could work though it will still only be 1 queue. A second queue was not preferred.

2. Q: Providing the symbol resolution in quark adds a fair bit of extra code, what to do?.
A: Removed all the specific code for doing symbol resolution, you can still use the tls_attach function with offsets and resolve symbols in userspace. The symbol related functions is easier to add in separate PR.

3. Q: How do we handle lingering map entries for all the connection ID's for a process that for some reason was not freed (like if the process crashes).
A: Moved the code for cleaning up on process exit into the TLS ebpf program itself, mirroring the way that Process/Probe.bpf.c does it, on last thread exiting.

4. Q: Where should I exclude tests from Valgrind, as Uprobes causes SIGTRAP?
A: Removed the Makefile entries and added them to the tests themself as other tests currently do.

## How to use it:
Now to explain a bit about how it works, there is 2 usecases that the public API exposes, the logic to determine **WHEN** to do this is up to either a future quark addition (a rule action as a pure example) or determined by the program that imports quark, as these are attached on-demand, then drained exactly like other events through `GetEvent`

### TlsAttach
When a binary you want to attach to compiles its own static version of `OpenSSL/BoringSSL` you provide the `pid`, `path` and the `offsets` to use (new, read, write, free). It is recommended that you use `-1` as the `pid`, and track this by `(dev, inode)`. This is because you do not need to apply a new tls attachment each time someone starts the same binary or use the same inode object, so when your logic triggers a second time indicating possible new tls attachment, you check if the inode object is already attached and skips, but it's left up to the user on what they want to do. Multiple uprobes on the same inode objecct (but scoped to their specific pid) is refcounted, so it will still only create 1 int3 breakpoints rather than multiple.
It's the same with preventing TOCTOU and similar is up to the user through various common techniques.

### TlsAttachSym (Removed from this PR, will be opened separately, resolve symbols yourself at first)
Mainly meant as a helper for attaching to system shared libraries like `libssl.so`. Currently it does this by already defining the required symbol names.
All you provide is the pid and the inode object path and restricts you from passing `-1` as the PID to prevent mistakes, this is because the overhead is not in filtering out the events but the fact that it adds a `int3` breakpoint for ANY process using `libssl`.

### Draining events
Draining the collected events is done through the same mechanism as before, your `GetEvent` now returns one of three new types that you branch on:

- QUARK_EV_TLS_CONN_ESTABLISHED (ConnId, PID, PrefixUnknown, EventTime)
- QUARK_EV_TLS_CALL (The actual event struct and payload)
- QUARK_EV_TLS_CONN_CLOSED (ConnId, PID, EventTime)

### Closing down

Both `TlsAttach` and `TlsAttachSym` returns a handler, whenever you want to close it again you call `TlsDetach` with the provided handler.

When shutting down, quark will automatically close any handler still open as part of its normal graceful shutdown in case something didn't go wrong, but preferably this would already be empty.

Currently a process with remaining references to open connections are automatically cleaned up on process exit (as mentioned in the open questions at the top).

## Event Flow Chart

```mermaid
flowchart TD
  A["quark_queue_tls_attach / _attach_sym (userspace)"] -->|"one bpf_link per program, stored in ta->links[]"| B["uprobes now live on libssl in the target"]
  B --> C{"target calls an SSL_* function"}
  C -->|SSL_new returns| N["uretprobe__ssl_new"]
  C -->|SSL_write entry+return| W["uprobe/uretprobe__ssl_write"]
  C -->|SSL_read entry+return| R["uprobe/uretprobe__ssl_read"]
  C -->|SSL_free entry| F["uprobe__ssl_free"]
  N -->|"reserve+submit (fixed size)"| RB[("shared ringbuf")]
  W -->|"state set → get → capture"| CAP["tls_capture_io"]
  R -->|"state set → get → capture"| CAP
  CAP -->|"tls_emit_call → bpf_loop → tls_emit_chunk"| PCPU["per-CPU event_buffer_map (scratch)"]
  PCPU -->|"ebpf_ringbuf_write = copy in"| RB
  F -->|reserve+submit| RB
  RB -->|"ring_buffer__consume_n, called INSIDE GetEvent"| CB["bpf_ringbuf_cb → ebpf_events_to_raw"]
  CB -->|raw_event_insert| T["by_time + by_pidtime RB-trees"]
  T -->|"held ~hold_time, then oldest (RB_MIN)"| AGG["quark_queue_aggregate: chains chunks of one call"]
  AGG -->|raw_event_tls_conn / raw_event_tls_call| ES["event_storage (one reused quark_event)"]
  ES -->|"GetEvent() returns it"| GO["your Go Event → your drain"]
```

### DIFF
--- a/elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
@@ -122,7 +122,6 @@ static long tls_emit_chunk(u32 idx, void *ctx)
     task               = (struct task_struct *)bpf_get_current_task();
     event->hdr.type    = EBPF_EVENT_TLS_CHUNK;
     event->hdr.ts      = bpf_ktime_get_ns();
-    event->hdr.ts_boot = bpf_ktime_get_boot_ns_helper();
     ebpf_pid_info__fill(&event->pids, task);
     event->conn_id      = cctx->conn_id;
     event->direction    = (uint8_t)cctx->direction;
@@ -205,7 +204,6 @@ static void tls_emit_conn(u64 conn_id, u32 flags)
     task               = (struct task_struct *)bpf_get_current_task();
     event->hdr.type    = EBPF_EVENT_TLS_CONN;
     event->hdr.ts      = bpf_ktime_get_ns();
-    event->hdr.ts_boot = bpf_ktime_get_boot_ns_helper();
     ebpf_pid_info__fill(&event->pids, task);
     event->conn_id = conn_id;
     event->flags   = flags;
@@ -322,7 +320,6 @@ int BPF_UPROBE(uprobe__ssl_free, void *ssl)
     task               = (struct task_struct *)bpf_get_current_task();
     event->hdr.type    = EBPF_EVENT_TLS_CONN_CLOSE;
     event->hdr.ts      = bpf_ktime_get_ns();
-    event->hdr.ts_boot = bpf_ktime_get_boot_ns_helper();
     ebpf_pid_info__fill(&event->pids, task);
     event->conn_id = conn_id;
 
--- a/elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
@@ -19,13 +19,7 @@
 
 // Allow-list of TGIDs to capture TLS from. Opposite of trusted_pids (a
 // deny-list): a TGID must be present here for a connection to be minted.
-// Populated by quark_queue_track_tgid()/untrack_tgid(). SSL_new consults it on
-// every call. SSL_read/SSL_write consult it only when they miss tls_conns, to
-// decide whether the miss is a tracked connection whose SSL_new was never seen
-// -- which must be adopted, not dropped -- or an untracked caller to ignore.
-// SSL_free needs no check: it only acts on an existing tls_conns entry, and
-// only gated paths ever create one. A tls_conns hit is therefore always
-// capture-worthy, so the capture hot path stays a single lookup.
+// Populated by quark_queue_track_tgid()/untrack_tgid().
 struct {
     __uint(type, BPF_MAP_TYPE_HASH);
     __uint(key_size, sizeof(u32));
@@ -39,11 +33,8 @@ static bool ebpf_events_is_tracked_tgid()
     return bpf_map_lookup_elem(&tracked_tgids, &tgid) != NULL;
 }
 
-// Global monotonic connection-id counter, bumped once per SSL_new. A single
-// array cell (not a per-tgid hash) so conn_id is unique across the whole
-// system -- consumers can key on conn_id alone -- and it is never reused, so a
-// closed-then-reopened SSL object at the same address gets a distinct id.
-// conn_id 0 is never issued, it is reserved as the "unset" sentinel.
+// Global monotonic connection-id counter, bumped once per SSL_new. conn_id 0 is
+// never issued, it is reserved as the "unset" sentinel.
 struct {
     __uint(type, BPF_MAP_TYPE_ARRAY);
     __uint(key_size, sizeof(u32));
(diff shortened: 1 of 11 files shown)


nicholasberlin:
Looking more myself, but the bot seems to have found some things worth investigating

----
# PR Review: #358 — Adding initial TLS attach/detach mechanism

**Author:** P1llus
**Branch:** tls-uprobe -> main
**Commits:** 20 | **Files changed:** 11 | **+2448 -21**
**URL:** https://github.com/elastic/quark/pull/358
**Reviewed commit:** 5bcc0ee1882c8064868f9d07d7468633ccabca3d
**Reviewed by:** GPT-5.5

## Summary

This PR adds TLS uprobe attachment APIs, TLS event wire formats, userspace aggregation, Go bindings, and a useful set of e2e tests. The core design looks coherent, but I found two correctness issues that should be addressed before merge: one can emit impossible connection lifecycle state under map pressure, and the public attach API can mis-handle queues that fell back away from the eBPF backend.

## Findings

### Significant

1. `SSL_new` can emit an ESTABLISHED event for a connection that was never inserted into `tls_conns`.

   In `elastic-ebpf/GPL/Events/TLS/Probe.bpf.c:278-282`, the `uretprobe__ssl_new` path checks that `tls_conn_tgids` was marked, but it does not check whether `bpf_map_update_elem(&tls_conns, ...)` actually succeeded before calling `tls_emit_conn()`. If `tls_conns` is full or the update fails for another reason, userspace will see `QUARK_EV_TLS_CONN_ESTABLISHED` with a fresh `conn_id`, but subsequent read/write/free probes cannot find that connection and no close event will be emitted. The late-adopt path already does the safer pattern by looking the key back up before announcing it; the normal `SSL_new` path should do the same or skip the establish on failed insert.

2. TLS attach can treat a non-eBPF fallback backend as `struct bpf_queue`.

   `quark_queue_open()` supports `QQ_ALL_BACKENDS` by trying eBPF first and falling back to kprobe if eBPF open fails (`quark.c:4247-4249`). After that fallback, `qq->flags` can still contain `QQ_EBPF`, but `qq->queue_be` points at a `struct kprobe_queue`. The new TLS APIs call `quark_get_bpf_probes()` and then use `qq->queue_be` as `struct bpf_queue` (`bpf_queue.c:1598-1606`, `bpf_queue.c:1807-1815`, `bpf_queue.c:1893-1903`); since `quark_get_bpf_probes()` only checks the requested flags, not `qq->stats.backend` or `qq->queue_ops`, a default `QQ_ALL_BACKENDS | QQ_TLS` queue that fell back to kprobe can dereference the wrong backend layout when the caller invokes `quark_queue_tls_attach*()`. The attach APIs should fail cleanly unless the active backend is eBPF, and `QQ_TLS` probably should either require eBPF at open time or disable fallback.

## Verdict

REQUEST_CHANGES. The feature is promising and well-tested for the happy paths, but the two lifecycle/backend correctness issues above can produce misleading events or unsafe API behavior in realistic failure/fallback scenarios.

P1llus:
Thanks @nicholasberlin . Yeah the last one is a good gap to cover, it's only meant to work with the EBPF backend either way.

The first one is also correct, its a good shoutout though its not really a "significant" finding it's something worth resolving.

I have kind of put this on hold as I am awaiting some answers on the open questions in my PR description, once that is done I'l fix those and do one last review. We are also doing some of the discussions outside the github issue which is not always good in terms of sharing context.

elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
```diff
@@ -0,0 +1,518 @@
...
+
+    off = idx * TLS_CHUNK_MAX;
+
+    event = get_event_buffer();
+    if (!event)
+        return 1;
+
+    task               = (struct task_struct *)bpf_get_current_task();
+    event->hdr.type    = EBPF_EVENT_TLS_CHUNK;
+    event->hdr.ts      = bpf_ktime_get_ns();
```
nicholasberlin, line 124:
```suggestion
    event->hdr.ts      = bpf_ktime_get_boot_ns();
```

elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
```diff
@@ -0,0 +1,518 @@
...
+    struct ebpf_tls_new_event *event;
+    struct task_struct *task;
+
+    event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+    if (!event)
+        return;
+
+    task               = (struct task_struct *)bpf_get_current_task();
+    event->hdr.type    = EBPF_EVENT_TLS_CONN;
+    event->hdr.ts      = bpf_ktime_get_ns();
```
nicholasberlin, line 206:
```suggestion
    event->hdr.ts      = bpf_ktime_get_boot_ns();
```

elastic-ebpf/GPL/Events/TLS/Probe.bpf.c
```diff
@@ -0,0 +1,518 @@
...
+    conn_id = conn->conn_id;
+    bpf_map_delete_elem(&tls_conns, &key);
+
+    event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+    if (!event)
+        goto out;
+
+    task               = (struct task_struct *)bpf_get_current_task();
+    event->hdr.type    = EBPF_EVENT_TLS_CONN_CLOSE;
+    event->hdr.ts      = bpf_ktime_get_ns();
```
nicholasberlin, line 322:
```suggestion
    event->hdr.ts      = bpf_ktime_get_boot_ns();
```

nicholasberlin (commented):
`bpf_ktime_get_ns` gets time since boot, but does not include suspend time. Where as `bpf_ktime_get_boot_ns` gets time since boot and does include suspend time.

`bpf_ktime_get_boot_ns` was added in 5.8, the same time as the bpf ring buffer. So support was always there, but we didn't realize it.

Anyway, always use `bpf_ktime_get_boot_ns` when getting times.

Issue #362, opened 2026-07-14
Add something like Archlinux to the CI


haesbaert:
Archlinux updates their kernel more often than fedora/ubuntu, so we can catch things earlier.

Just spent sometime debugging why things didn't work on cachyos, which is archlinux with a gentoo spirit. Basically they enable LTO which causes some symbols to be inlined, like vfs_write() and then bad things happen.

We should still add Archlinux since it's "the same kernel" but without LTO:
```
zcat /proc/config.gz |grep LTO
CONFIG_ARCH_SUPPORTS_LTO_CLANG=y
CONFIG_ARCH_SUPPORTS_LTO_CLANG_THIN=y
CONFIG_LTO_NONE=y
CONFIG_HID_WALTOP=m
```

This should be fairly easy, it's just adapting a `krun-arch.sh`.

haesbaert:
cc @gogochan @dwhyrock

PR #363, opened 2026-07-14
Add archlinux to krun-archlinux.sh and CI

haesbaert:
Pretty straightforward, same dance as ubuntu and fedora, except there's no
version since it's always rolling.

Archlinux updates kernels faster than ubuntu and fedora, so we might catch
problems earlier.

Prompted by an issue with cachyos where quark-test(8) will fail due to them
using LTO in the kernel and inlinining vfs_write() calls.

Users of cachyos can use the kernel distributed by archlinux, which does not
have LTO and therefore our probes are dandy.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -512,3 +512,14 @@ steps:
       provider: gcp
       machineType: n2-standard-2
       enableNestedVirtualization: true
+
+  - label: "quark-test on archlinux (rolling)"
+    key: test_archlinux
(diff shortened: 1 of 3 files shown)


haesbaert:
related to #362

haesbaert:
cc @AndersonQ @belimawr @pchila @WiegerElastic

belimawr (commented):
LGTM, thanks for adding Arch to the test matrix!

nicholasberlin (approved):

PR #364, opened 2026-07-14
Prefer boottime timestamps for wallclock event stamping

nicholasberlin:
The bpf backend stamps `raw->time` from the probes' `ts` field, which is `bpf_ktime_get_ns()` (`CLOCK_MONOTONIC`) and stops during suspend. The boot time epoch it is later added to includes suspended time, so after a suspend every event timestamp falls behind the wall clock by the total suspended time until reboot. Process start times do not have this problem since task `start_boottime` is `CLOCK_BOOTTIME`.

The probes already capture `ts_boot` (`bpf_ktime_get_boot_ns()`, zero on kernels older than 5.7). Carry it on the raw event and prefer it when converting to a wall clock timestamp in `quark_queue_get_event()` and for exit_time_event. `raw->time` stays monotonic since internal ordering and aging compare it against `now64()`, which is `CLOCK_MONOTONIC`.

### DIFF
--- a/bpf_queue.c
+++ b/bpf_queue.c
@@ -172,7 +172,8 @@ ebpf_events_to_raw(struct quark_queue *qq, struct ebpf_event_header *ev)
 		}
 
 		raw->task.exit_code = exit->exit_code;
-		raw->task.exit_time_event = raw->time;
+		raw->task.exit_time_event =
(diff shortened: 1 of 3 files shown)


nicholasberlin:
Closing in favor of this https://github.com/elastic/quark/pull/365

PR #365, opened 2026-07-16
Prefer CLOCK_BOOTTIME over CLOCK_MONOTONIC

nicholasberlin:
The probes stamped events with bpf_ktime_get_ns(), which is
CLOCK_MONOTONIC and stops during suspend, so every emitted timestamp
fell behind the wallclock by the total suspended time until reboot.
Process start times never had this problem since task start_boottime
is CLOCK_BOOTTIME, meaning event times and start times did not even
share a domain.

The kernel test matrix confirmed every supported bpf configuration has
bpf_ktime_get_boot_ns (in the kernel since 5.8), so the probes now
stamp ts with it directly, and the separate ts_boot field along with
its CO-RE existence check helper are gone. The helper call is
unconditional: on kernels without it the programs fail to load,
bpf_queue_open() fails and quark falls back to the kprobe backend.

The kprobe backend is the only one left with monotonic times, as perf
stamps samples with the monotonic-ish sched clock. Introduce
QQ_MONOTONIC on the queue flags to relay this: informational only, set
by the kprobe backend on the opened queue, refused with EINVAL if
given as configuration. Monotonic times undercount by the total
suspended time, users that care about suspend must compensate.

now64() moves to the queue's time domain (CLOCK_BOOTTIME, or
CLOCK_MONOTONIC on QQ_MONOTONIC queues) so aggregation ordering and
aging always compare event times against a clock in the same domain.

Add t_monotonic to quark-test for all backends, checking the flag is
refused as configuration and relays the backend's domain after open.
Document QQ_MONOTONIC in quark_queue_open.3.

### DIFF
--- a/docs/quark_queue_open.3.html
+++ b/docs/quark_queue_open.3.html
@@ -89,6 +89,13 @@
       <dd>Enable the KPROBE backend, see above.</dd>
       <dt id="QQ_ALL_BACKENDS"><a class="permalink" href="#QQ_ALL_BACKENDS"><code class="Dv">QQ_ALL_BACKENDS</code></a></dt>
       <dd>Shorthand for (QQ_EBPF | QQ_KPROBE).</dd>
+      <dt id="QQ_MONOTONIC"><a class="permalink" href="#QQ_MONOTONIC"><code class="Dv">QQ_MONOTONIC</code></a></dt>
+      <dd>Informational, not configuration: specifying it fails with
+          <code class="Er">EINVAL</code>. The backend that opens sets it on the
+          opened queue's <i class="Em">flags</i> when its event times are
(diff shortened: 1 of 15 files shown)


haesbaert:
I like this, I wanna go over it a bit more carefully but I think this is the right direction.
One minor neat, can we move the time_monotonic to QQ_MONOTONIC inside flags, it can even double as a config option. If the kernel doesn't support, you can just erase the flags, same functionality, just avoids poluting the main structure with a bool flag.

nicholasberlin:
> move the time_monotonic to QQ_MONOTONIC inside flags

done, new docs

haesbaert (approved):

PR #368, opened 2026-07-17
Remove boottime from quark times

nicholasberlin:
Times derived from quark.boottime were computed once and stored, so a system clock step after quark_init() (say NTP correcting a clock that was wrong at boot) left every stored and emitted timestamp shifted by the step size until restart.

Nothing computed with quark.boottime is stored or handed out anymore: qev->time, proc_time_boot and exit_time_event are raw ns since boot, translated at the last moment with the new quark_time_to_wallclock(). Since quark has no event loop, keeping the epoch honest is the user's job: detect a step (a CLOCK_REALTIME timerfd with TFD_TIMER_CANCEL_ON_SET, or polling the CLOCK_REALTIME - CLOCK_BOOTTIME delta, which only moves on steps) and call the new quark_update_boottime(). Corrections then apply retroactively to any raw time, including events that predate the step.

This also makes proc_entity_id immune to clock changes, as it now hashes the raw start time.

Breaking change: consumers that treated qev->time, proc_time_boot or exit_time_event as wallclock epoch ns must translate via quark_time_to_wallclock().

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -123,3 +123,16 @@ Changes from 0.6 to 0.7
     system.
   o The quark utilities now display their manpage when invoked with
     "help", "-help" or "--help".
+
+Changes from 0.7 to 0.8
+~~~~~~~~~~~~~~~~~~~~~~~
+  o BREAKING, SILENTLY: quark_process.proc_time_boot,
+    quark_process.exit_time_event and quark_event.time are now
+    expressed in nanoseconds since boot instead of nanoseconds since
+    the epoch. Existing code compiles unchanged but any user treating
+    these fields as wallclock will emit timestamps that are off by
+    the boot time. Users that want wallclock time must convert with
+    the new quark_time_to_wallclock(). quark_get_boottime() and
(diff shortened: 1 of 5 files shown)


haesbaert:
heya, sorry for the delay.
This needs some reasonable entry in CHANGES, since it's an actual breaking change. I'm trying to prioritize things that are blocking others but I'll get back to this.

haesbaert:
but change looks good, I see nothing wrong with it, might make sense to coordinate with @biscout42 just so that it doesn't break his prototype.

biscout42:
No need to coordinate ;) and feel free to merge this one. We will update PoC, if necessary.

haesbaert:
> No need to coordinate ;) and feel free to merge this one. We will update PoC, if necessary.

theeeeeeennnnnn f**k it, we ball :)

haesbaert (approved):

haesbaert:
I approved just make sure to add the entry to CHANGES and feel free to merge it.

PR #369, opened 2026-07-23
Add file.exec_change rule predicate

mjwolf:
Adds a new rule predicate which matches when a file's execute bit is set as a create or a permission change. It does not match on file content changes to already-executable files.

This also adds a file change mask, which will propagate the file change reasons from ebpf events to aggregated quark events.

This is not supported in the nova backend, and will fail with ENOTSUP, if there's an attempt to use this rule predicate with that backend.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -295,7 +295,7 @@ steps:
 
   - label: "quark-test on rhel 8.5 (no file)"
     key: test_rhel_8_5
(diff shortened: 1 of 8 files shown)


haesbaert (approved):
This looks good!
I'm not sure if it shouldn't be generic, like "file attribute changed" for the matching field, but if so yall can add it later, there's no point in gating this.

haesbaert:
I fixed CI, file tests have to be escaped on ancient rhel 8.x.

PR #371, opened 2026-07-31
Don't return NULL unless we have no events

haesbaert:
We would incorrectly return NULL on quark_queue_get_event() when an event was
dropped, this would signal the caller that "there are no more events" and it
should quark_queue_block().

This makes sure we try to pop the next event in case we get a RA_DROP rule.

Very likely my last commit, as today is my last day at elastic.

So Long, and Thanks for All the Fish.

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -4563,12 +4563,6 @@ quark_queue_pop_raw(struct quark_queue *qq)
 	struct raw_event	*min;
 	u64			 now;
 
-	/*
-	 * We populate before draining so we can have a fuller tree for
-	 * aggregation.
-	 */
-	(void)quark_queue_populate(qq);
-
 	now = now64(qq);
 	min = RB_MIN(raw_event_by_time, &qq->raw_event_by_time);
 	if (min == NULL || !raw_event_expired(qq, min, now)) {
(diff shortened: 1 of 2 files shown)


quark.c
```diff
@@ -5200,11 +5182,39 @@ quark_queue_get_event(struct quark_queue *qq)
...
+	if (qq->flags & QQ_BYPASS)
+		return (get_bypass_event(qq));
+
+	/*
+	 * Populate once before draining so we can have a fuller tree for
+	 * aggregation without chasing a continuously replenished stream.
+	 */
+	(void)quark_queue_populate(qq);
+
+next:
```
biscout42, line 5211: nit: totally up-to you folks. Would a loop better here from readability perspective?

quark.c
```diff
@@ -5200,11 +5182,39 @@ quark_queue_get_event(struct quark_queue *qq)
...
+	qev = quark_queue_get_event1(qq);
+	/* Run ruleset, if we drop, get the next event */
+	if (qev != NULL && qq->ruleset != NULL) {
```
biscout42, line 5214: Do I understand it correct, that if all the events in the buffer are filtered by a ruleset, we will return `NULL` even if there are events in the ring buffer? IMHO the PR doesn't change the today's situation, but reduces the frequency of unnecessary `wait` signals to the API caller.

biscout42 (approved):
Could we add unit tests to verify, that a batch with an event filtered by a ruleset doesn't return `NULL` on a first event? And one for the issue the latest commit fixes?

matthewscherer (approved):

quark.c
nicholasberlin: i prefer to leave it, as there are many other instances like this in the code base.

quark.c
nicholasberlin: Yeah that's right ... if the entire populated batch is dropped we still return NULL even if newer events sit in the ring buffer.

The events in the ring buffer arrived less than `hold_time` ago, so they wouldn't be deliverable anyway.

The next call will pick them up. The PR only stops us from signaling 'no events' while an acceptable event is already buffered.

Added tests. Both tests failed against quark.c previous to this PR.

biscout42: Thank you.

PR #372, opened 2026-08-03
Document boottime calls, add Go wrappers

nicholasberlin:
Follow-up to #368 (now merged, rebased onto main), closing the gaps noted there:

- Manpages for quark_update_boottime(3), quark_get_boottime(3) and
  quark_time_to_wallclock(3), listed in quark(7), with the HTML docs
  regenerated. The pages document which times are expressed since boot,
  that btime comes from /proc/stat and moves on clock steps, and two
  common ways of detecting a step (polling the CLOCK_REALTIME -
  CLOCK_BOOTTIME delta, or a timerfd armed with
  TFD_TIMER_CANCEL_ON_SET). Each page carries a HISTORY section noting
  the calls appeared in quark 0.8, when all times handed out by quark
  switched from epoch to since-boot, so readers on older versions
  aren't misled.
- The in-tree Go bindings (go/quark) learn UpdateBoottime(), Boottime()
  and TimeToWallclock(), mirroring the C calls, plus a non-root test
  (passes locally; TestQuark still requires root).

Also:

- The 0.8 CHANGES entry now lists quark_socket established_time and
  close_time in the BREAKING, SILENTLY bullet, which had missed them,
  and gains a bullet for the new Go calls.
- README.md regenerated via make README.md (html2markdown).

mandoc -Tlint passes on the new pages, go vet and gofmt are clean.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -127,7 +127,8 @@ Changes from 0.6 to 0.7
 Changes from 0.7 to 0.8
 ~~~~~~~~~~~~~~~~~~~~~~~
   o BREAKING, SILENTLY: quark_process.proc_time_boot,
-    quark_process.exit_time_event and quark_event.time are now
+    quark_process.exit_time_event, quark_event.time,
+    quark_socket.established_time and quark_socket.close_time are now
     expressed in nanoseconds since boot instead of nanoseconds since
     the epoch. Existing code compiles unchanged but any user treating
     these fields as wallclock will emit timestamps that are off by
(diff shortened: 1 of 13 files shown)


nicholasberlin:
Superseded by single-concern PRs, one per concern:

- #387 — the CHANGES fix (socket times in the 0.8 breaking bullet)
- #388 — the manpages, with README/docs regenerated
- #389 — the Go boottime wrappers (UpdateBoottime, Boottime, TimeToWallclock)

The code in those PRs is identical to this one.

PR #376, opened 2026-08-07
Add process_vm_access events for process_vm_readv/writev

Aegrah:
## Summary

Emit `process_vm_access` when a process uses `process_vm_readv` or `process_vm_writev` to access another process's memory. Events carry the operation, target PID, first remote address, requested byte count, and raw syscall return value so Defend can distinguish successful transfers from failed attempts. Same-process calls are suppressed.

Issue: https://github.com/elastic/endpoint-dev/issues/21204

- Hook/backend: paired `sys_enter/sys_exit_process_vm_{readv,writev}` tracepoints, eBPF-only
- Completion/state: bounded per-thread state correlates entry arguments with the raw syscall return value at exit
- Event/flag: `QUARK_EV_PROCESS_VM_ACCESS`, `QQ_PROCESS_VM_ACCESS`
- Compatibility: same-process calls are suppressed; successful and failed cross-process calls are emitted; no kprobe fallback

## Proof

- Build: PASS (`make` + clang-17)
- Sensor test: `t_process_vm_access` PASS (read/write events present; same-process calls absent)
- Cold Fleet acceptance: `20260731T222145Z-process-vm-access` (9.4.2)
- Persona volume gate: `20260802T092603Z-process-vm-access-volume` on `endpoint-9.4.2-ac8c21cfbdbcac3d` (PASS, 4h; 0 target docs/h as expected)


christos68k:
Hi @Aegrah, is this PR ready for review?

Aegrah:
Hi @christos68k; thanks for checking. I'm still working with @nicholasberlin on some of the fundamentals. Once it's ready, I will mark it for review!

Aegrah:
Superseded by #416.

That PR resolves the target from the task `mm_access()` has already looked up, rather
than from the `pid` argument at syscall entry — which is namespace-relative and may name
a thread, so it was being compared against a global tgid. It also adds
`target_start_time_ns` so a reused PID cannot be mistaken for the original target.

Closing in favour of #416.

PR #377, opened 2026-08-11
Do not aggregate away executable changes

nicholasberlin:
Keep file events separate when their executable state changes, so a
`file.exec_change` rule cannot miss a transient executable mode
(e.g. `chmod +x` immediately followed by `chmod -x`). Previously the
two chmods aggregated into a single event whose final mode had no
exec bit, and the rule never fired.

Adds regression coverage to `t_rule_exec_change` for the transient
permission sequence.

Verified in the QEMU test rig (kernel 6.8.0-65-generic):
- Without the `quark.c` change, the new test fails (`drain_for_pid: Timer expired`).
- With it, the full suite passes (0 failures).


Tacklebox (approved):
Great catch 👍

PR #380, opened 2026-08-12
Split pods/containers into separate trees in quark_queue

Tacklebox:
Pods and containers need to be accessible independently of Kubernetes state. This change moves the pod index from `quark_kube` to `quark_queue`, alongside the existing container index, and adds public pod/container get and lookup helpers.

`quark_container_get()` supports creating a container without a pod and attaching it later. Repeated calls preserve the existing object; requesting a different parent for an attached container fails with `EEXIST`. Queue shutdown frees all pods, their attached containers, and any remaining orphan containers.

Independent container garbage collection and public pod/container removal APIs are implemented in follow-up [PR #383](https://github.com/elastic/quark/pull/383). That PR adds `GC_CONTAINER`, container GC queue linkage and cleanup, and C/Go removal helpers that schedule collection after the queue's cache grace time. In this PR, containers attached to a pod are freed with that pod; orphan containers are freed when the queue closes.

Validation on the local checkout, including the review fixes:

- `make ARCH=arm64 docker` passed using the repository Dockerfile.
- `t_pod_get` and `t_container_get` passed in the Linux builder container.
- `git diff --check` passed.

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -836,11 +836,10 @@ pod_by_uid_cmp(struct quark_pod *a, struct quark_pod *b)
 static struct quark_pod *
 pod_lookup_by_uid(struct quark_queue *qq, char *uid)
 {
-	struct quark_kube	*qkube = qq->qkube;
 	struct quark_pod	 key, *k;
 
 	key.uid = uid;
-	k = RB_FIND(pod_by_uid, &qkube->pod_by_uid, &key);
+	k = RB_FIND(pod_by_uid, &qq->pod_by_uid, &key);
 	if (k == NULL)
 		errno = ESRCH;
 
@@ -850,7 +849,6 @@ pod_lookup_by_uid(struct quark_queue *qq, char *uid)
 static int
 pod_insert(struct quark_queue *qq, struct quark_pod *pod)
 {
-	struct quark_kube	*qkube = qq->qkube;
 	struct quark_pod	*col;
 
 	if (pod->linked) {
@@ -858,7 +856,7 @@ pod_insert(struct quark_queue *qq, struct quark_pod *pod)
 		return (-1);
 	}
 
-	col = RB_INSERT(pod_by_uid, &qkube->pod_by_uid, pod);
+	col = RB_INSERT(pod_by_uid, &qq->pod_by_uid, pod);
 	if (unlikely(col != NULL))
 		return (errno = EEXIST, -1);
 
@@ -870,12 +868,11 @@ pod_insert(struct quark_queue *qq, struct quark_pod *pod)
 static void
 pod_delete(struct quark_queue *qq, struct quark_pod *pod)
 {
-	struct quark_kube	*qkube = qq->qkube;
 	struct quark_container	*container;
 	struct label_node	*node;
 
 	if (pod->linked) {
-		RB_REMOVE(pod_by_uid, &qkube->pod_by_uid, pod);
+		RB_REMOVE(pod_by_uid, &qq->pod_by_uid, pod);
 		pod->linked = 0;
 	}
 	gc_unlink(qq, &pod->gc);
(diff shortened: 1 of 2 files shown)


nicholasberlin:
# PR Review: #380 — Split pods/containers into separate trees in quark_queue

**Author:** Tacklebox
**Branch:** oci-trees-structure -> main
**Commits:** 1 | **Files changed:** 2 | **+116 -11**
**URL:** https://github.com/elastic/quark/pull/380
**Reviewed commit:** 499635fd183c061839e924c89c1ed4efac0231f7
**Reviewed by:** OpenAI GPT-5

## Summary

This change moves the pod index onto `quark_queue`, makes pod/container lookup available without Kubernetes state, adds public get/lookup helpers, and cleans up queue-owned pods and orphan containers on close. The ownership move itself is coherent, but the new container getter does not honor its documented parent-link behavior for existing containers, and the reviewed commit does not contain the independent container-GC work claimed by the PR description.

## Findings

### Significant

1. **Existing containers are returned without applying or validating `pod_uid`** (`quark.c:2402`).

   `quark_container_get()` resolves the requested pod, but then immediately returns when `container_lookup()` finds an existing container. A supported call sequence such as `quark_container_get(qq, "id", NULL)` followed later by `quark_container_get(qq, "id", "pod-uid")` therefore leaves the container orphaned forever, despite the function's contract saying that a non-NULL `pod_uid` links the container to that pod. The same path silently returns a container already linked to a different pod. Because there is no other public attachment operation, callers cannot repair either state. When an existing container has no parent, link it into the requested pod; when it already belongs to a different pod, return a defined error such as `EEXIST` or `EINVAL`.

2. **The commit does not implement the independent container garbage collection described as part of this PR** (`quark.h:571`, `quark.h:695`, `quark.c:473`, `quark.c:789`).

   At the reviewed head, `enum gc_type` has no `GC_CONTAINER`, `struct quark_container` has no leading `gc_link`, `gc_collect()` has no container case, and `container_delete()` cannot unlink a queued container. Those are all explicitly listed in the PR summary as the structural foundation being delivered. They currently appear only in a later draft PR in the stack. Either include that foundation here or update this PR's scope and description so reviewers and merge sequencing reflect what this commit actually provides.

### Minor

1. **The comments still describe the moved trees as belonging to `quark_kube`** (`quark.h:712`, `quark.h:736`).

   After this change both global indexes live in `quark_queue`, so these comments now document the wrong owner and preserve Kubernetes-specific naming for a queue-wide data structure.

## Verdict

**REQUEST_CHANGES** — fix the existing-container parent-link path before exposing `quark_container_get()`, and reconcile the implementation with the PR's stated container-GC scope. The ownership relocation and close-time cleanup otherwise look sound.

## Verification

- `git diff --check origin/main...HEAD` passed.
- GitHub reports successful `buildkite/quark` and CLA statuses for the reviewed commit.
- `make libquark_big.a` could not be validated locally on macOS because vendored `libelf` fails first on Clang signedness warnings promoted to errors; the identical failure occurs at the base commit `d795343d147d0120b7c7a8e59b76a833f4b8b812`, before compiling the PR's changed files.

quark.c
```diff
@@ -4402,11 +4497,15 @@ quark_queue_close(struct quark_queue *qq)
 	/* Clean up backend */
 	if (qq->queue_ops != NULL)
 		qq->queue_ops->close(qq);
+	/* Clean up all pods (transitively cleans up their containers) */
+	while ((pod = RB_ROOT(&qq->pod_by_uid)) != NULL)
+		pod_delete(qq, pod);
+	/* Clean up any orphaned containers not linked to a pod */
+	while ((container = RB_ROOT(&qq->container_by_id)) != NULL)
```
biscout42, line 4504: Do we actually need it? IIUC `pod_delete` handles this case at https://github.com/elastic/quark/blob/main/quark.c#L891

quark.c
```diff
@@ -2344,6 +2341,104 @@ quark_process_lookup(struct quark_queue *qq, int pid)
...
+	struct quark_container	*container, *col;
+	struct quark_pod	*pod = NULL;
+
+	if (pod_uid != NULL) {
+		pod = pod_lookup_by_uid(qq, (char *)pod_uid);
+		if (pod == NULL)
+			return (NULL);
+	}
+
+	container = container_lookup(qq, (char *)container_id);
```
biscout42, line 2402: Could we use `quark_container_get` with `pod_uid` being `NULL` ? If so, I assume we will not be able to attach a `pod_uid`  later, because of the check in the next line. Is that correct understanding? (the first finding from Nick's review)

Could we add a unit test which verifies and demoes correct use case of the function?

quark.c
```diff
@@ -2344,6 +2341,104 @@ quark_process_lookup(struct quark_queue *qq, int pid)
 	return (process_cache_get(qq, pid, 0));
 }
 
+/*
+ * Get or create a pod by uid. Returns the existing pod if already present,
+ * otherwise allocates and inserts a minimal pod (uid set, containers/labels
+ * initialized). Caller fills in remaining fields (name, ns, phase, etc.).
+ */
+struct quark_pod *
+quark_pod_get(struct quark_queue *qq, const char *uid)
```
biscout42, line 2350: Would it make sense to add a unit test?

biscout42 (changes requested):

quark.c
biscout42: The orphan cleanup is necessary: pod_delete() only frees containers attached to that pod, while the new API permits containers without a pod.

biscout42 (approved):
@nicholasberlin , I have addressed all the comments and changed PR description (because GC is implemented in the follow-up PR). 

Could you double-check the changes?

PR #381, opened 2026-08-12
Add functions to insert pods/containers to the new trees

Tacklebox:
Add C functions to create pods and containers in the queue.
Copy the supplied values into each new object.
Permit optional fields to be NULL.
Permit containers without a pod.

Return NULL and set errno to EEXIST for a duplicate ID.
Return NULL and set errno to ESRCH if the requested pod is missing.

Address the code review comments.
Print `<unknown>` for missing pod and container fields in event output.
Add creation, lookup, and error tests.
Add event output tests for missing fields and supplied values.

The Docker build passed with `make ARCH=arm64 docker`.
All six pod and container tests passed on Linux.
The diff check passed.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -2439,6 +2439,100 @@ quark_container_lookup(struct quark_queue *qq, const char *container_id)
 	return (container_lookup(qq, (char *)container_id));
 }
 
+struct quark_pod *
+quark_pod_create(struct quark_queue *qq, const char *uid,
+    const char *name, const char *ns, const char *phase)
+{
+	struct quark_pod	*pod;
+
+	if (pod_lookup_by_uid(qq, (char *)uid) != NULL)
+		return (errno = EEXIST, NULL);
+
+	pod = calloc(1, sizeof(*pod));
+	if (pod == NULL)
+		return (NULL);
+	RB_INIT(&pod->containers);
+	RB_INIT(&pod->labels);
+	pod->uid = strdup(uid);
+	if (pod->uid == NULL)
+		goto fail;
+	if (name != NULL && (pod->name = strdup(name)) == NULL)
+		goto fail;
+	if (ns != NULL && (pod->ns = strdup(ns)) == NULL)
+		goto fail;
+	if (phase != NULL && (pod->phase = strdup(phase)) == NULL)
+		goto fail;
+	if (pod_insert(qq, pod) == -1)
+		goto fail;
+
+	return (pod);
+fail:
+	free(pod->uid);
+	free(pod->name);
+	free(pod->ns);
+	free(pod->phase);
+	free(pod);
+	return (NULL);
+}
+
+struct quark_container *
+quark_container_create(struct quark_queue *qq, const char *container_id,
+    const char *pod_uid, const char *name, const char *image)
+{
+	struct quark_container	*container, *col;
+	struct quark_pod	*pod = NULL;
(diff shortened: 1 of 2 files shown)


quark.c
```diff
@@ -2439,6 +2439,100 @@ quark_container_lookup(struct quark_queue *qq, const char *container_id)
...
+quark_container_create(struct quark_queue *qq, const char *container_id,
+    const char *pod_uid, const char *name, const char *image)
+{
+	struct quark_container	*container, *col;
+	struct quark_pod	*pod = NULL;
+
+	if (pod_uid != NULL) {
+		pod = pod_lookup_by_uid(qq, (char *)pod_uid);
+		if (pod == NULL)
+			return (NULL);
```
biscout42, line 2488: Question: how do we usually report errors in C? IMHO this case is kind of `pod not found` error. WDYT?

quark.c
```diff
@@ -2439,6 +2439,100 @@ quark_container_lookup(struct quark_queue *qq, const char *container_id)
...
+
+	pod = calloc(1, sizeof(*pod));
+	if (pod == NULL)
+		return (NULL);
+	RB_INIT(&pod->containers);
+	RB_INIT(&pod->labels);
+	pod->uid = strdup(uid);
+	if (pod->uid == NULL)
+		goto fail;
+	if (name != NULL && (pod->name = strdup(name)) == NULL)
```
biscout42, line 2459: Bot says, if we pass `name, `ns` or `pahse` as `NULL`, we will have a pod inserted into the tree, which will fail `quark_event_dump()`.

Would it make sense to add a guard and reject inserting null values?

quark.c
biscout42, line 1: Do we have a uni test for `quark_pod_create`? We could do a round trip test with create / get

quark.c
```diff
@@ -2439,6 +2439,100 @@ quark_container_lookup(struct quark_queue *qq, const char *container_id)
...
+	free(pod->uid);
+	free(pod->name);
+	free(pod->ns);
+	free(pod->phase);
+	free(pod);
+	return (NULL);
+}
+
+struct quark_container *
+quark_container_create(struct quark_queue *qq, const char *container_id,
```
biscout42, line 2479: Do we have a uni test for `quark_container_create`? We could do a round trip test with create / get

quark.c
```diff
@@ -2439,6 +2439,100 @@ quark_container_lookup(struct quark_queue *qq, const char *container_id)
...
+		return (errno = EEXIST, NULL);
+
+	container = calloc(1, sizeof(*container));
+	if (container == NULL)
+		return (NULL);
+	TAILQ_INIT(&container->processes);
+	container->container_id = strdup(container_id);
+	if (container->container_id == NULL)
+		goto fail;
+	if (name != NULL && (container->name = strdup(name)) == NULL)
```
biscout42, line 2501: same here, bot says `quark_event_dump()` will fail with this fields being `NULL`. Shall we prevent it?

biscout42 (changes requested):
My main ask if we have unit tests or could we add them?

biscout42 (dismissed):

PR #383, opened 2026-08-12
Exposed OCI trees and helpers to go-quark

Tacklebox:
Add Go functions to create, find, and remove pods and containers.
Include container and pod data in Go process events.

Add independent container garbage collection.
Collect removed objects after the cache grace time.
Remove child containers when their pod is collected.
Clear process links when a container is freed.

Validation on the local checkout:

- The default Docker build passed.
- Both Go snapshot and removal tests passed.
- All eight C pod and container tests passed.
- Both changed C files compiled with the CentOS 7 warning settings.
- The Go package built without the test backend.
- The diff check passed.

For now, we are reading same pod / container several times per an event, e.g. while creating an ecs document. If we see it creates a performance bottleneck, we will optimise (e.g. add memoization or keep a separate pod / container cache on a consumer side).

### DIFF
--- a/go/quark/quark.go
+++ b/go/quark/quark.go
@@ -94,16 +94,38 @@ type Exit struct {
 	Valid           bool
 }
 
+// PodInfo holds the fields of a Kubernetes pod.
+type PodInfo struct {
+	UID   string
+	Name  string
+	NS    string
+	Phase string
+}
+
+// ContainerInfo holds the fields of a container, with an optional link to its
+// parent pod.
+type ContainerInfo struct {
+	ContainerID string
+	Name        string
+	Image       string
+	ImageID     string
+	ImageName   string
+	ImageTag    string
+	ImageHash   string
+	Pod         *PodInfo // nil if the container has no pod link
+}
+
 // Process represents a single process.
 type Process struct {
-	Pid     uint32 // Always present
-	Proc    Proc   // Only meaningful if Proc.Valid (QUARK_F_PROC)
-	Exit    Exit   // Only meaningful if Exit.Valid (QUARK_F_EXIT)
-	Comm    string
-	Exe     string
-	Cmdline []string
-	Cwd     string
-	Cgroup  string
+	Pid       uint32 // Always present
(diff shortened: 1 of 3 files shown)


go/quark/quark.go
```diff
@@ -462,6 +484,49 @@ func SetVerbose(level int) {
 	C.quark_verbose = C.int(level)
 }
 
+func goStringMaybe(s *C.char) string {
```
biscout42, line 487: Do we really need it? I believe ` C.GoString(nil) returns ""`. Could you test it?

quark.c
```diff
@@ -2470,6 +2474,18 @@ quark_container_create(struct quark_queue *qq, const char *container_id,
 	return (NULL);
 }
 
+void
+quark_pod_remove(struct quark_queue *qq, struct quark_pod *pod)
```
biscout42, line 2478: Do we have unit tests for both functions - `quark_pod_remove` and `quark_container_remove`?

go/quark/quark.go
```diff
@@ -642,6 +711,115 @@ func shmFromC(cShm *C.struct_quark_shm) (any, error) {
 	return nil, fmt.Errorf("invalid shm kind")
 }
 
+// Pod is an opaque handle to a C quark_pod. Do not copy.
```
biscout42, line 714: nit: could we move type declaration on top of the file?

go/quark/quark.go
```diff
@@ -642,6 +711,115 @@ func shmFromC(cShm *C.struct_quark_shm) (any, error) {
...
+type Pod struct {
+	pod *C.struct_quark_pod
```
biscout42, line 716: IMHO C struct pointers here and in `Container` will bring us to troubles. Every `GetEvent` call frees internal C memory, and I assume this references become incorrect with sequential calls. 

Could we test it? If it is complex, I would suggest to change structure and pass only raw values, e.g. uuids and do lookup before delete in C code.

biscout42 (changes requested):
Overall looks good for me. I have one important question about C struct pointers in Pod / Container types.

biscout42 (dismissed):

PR #385, opened 2026-08-12
kprobe: don't exit when perf record gt 4096 bytes

nicholasberlin:
## Problem

`perf_mmap_read()` linearized a record that wraps the ring into a fixed `wrapped_event_buf[4096]` and `errx(1)`'d the whole process whenever `header.size` exceeded it. A single perf record can legally be bigger: `header.size` is a u16, and the `sched_process_exec` tracepoint records the full exec path (up to `PATH_MAX`) via `PERF_SAMPLE_RAW`. Execing a binary through a path longer than ~4KB therefore killed any libquark consumer on the kprobe backend:

```
quark-test: getting an event larger than wrapped buf
```

## Fix

Replace the fixed per-cpu buffer with one heap buffer per queue of `UINT16_MAX + 1` bytes: no record can exceed it by type, and the scratch is never live across two rings since each record is fully consumed before the next ring is read.

While here, harden the read path:
- Snapshot `header.size` once instead of re-reading it from the shared writable mapping.
- Reject corrupt records smaller than a header — consuming them would spin on the same offset forever.
- `static_assert` that the ring can hold the largest possible record, so a corrupt `data_head` can't walk the wrap copy out of the mapping.
- Drop the now-unused `err.h`.

Net effect is stronger than the old guard: even a fully corrupt `data_head` can no longer cause an out-of-bounds read or write — worst case is a garbage event, not heap corruption or a dead process.

## Testing

New `t_exec_long_path` (kprobe) execs a copy of `true` through a `PATH_MAX - 1` path and asserts the recorded `exe` matches byte-for-byte, proving the oversized record survives and is linearized intact.

- Against the previous code, the test reproduces the fatal exit (`getting an event larger than wrapped buf`, exit 1).
- With the fix, the full quark-test suite passes on both kprobe and ebpf backends (qemu, 6.8.0-65-generic).

### DIFF
--- a/kprobe_queue.c
+++ b/kprobe_queue.c
@@ -27,12 +27,13 @@
 #include "quark.h"
 
 #define PERF_MMAP_PAGES		16		/* Must be power of 2 */
+#define PERF_RECORD_MAX_SIZE	(UINT16_MAX + 1)	/* header.size is u16 */
 /*
  * A perf record caps at UINT16_MAX (perf_event_header.size). The ring must be
  * able to hold at least one such record, or perf_mmap_read() could chase a
  * corrupt data_head past the end of the ring. getpagesize() is at least 4096.
  */
-static_assert(PERF_MMAP_PAGES * 4096 >= UINT16_MAX + 1,
+static_assert(PERF_MMAP_PAGES * 4096 >= PERF_RECORD_MAX_SIZE,
     "perf ring can't hold the largest possible perf record");
 
 struct perf_sample_id {
@@ -118,6 +119,7 @@ struct perf_mmap {
 	size_t				 data_mask;
 	u8				*data_start;
 	u64				 data_tmp_tail;
+	int				 stalled;
 };
 
 struct perf_group_leader {
@@ -642,6 +644,7 @@ perf_mmap_init(struct perf_mmap *mm, int fd)
 	mm->data_mask = mm->data_size - 1;
 	mm->data_start = (uint8_t *)mm->metadata + getpagesize();
 	mm->data_tmp_tail = mm->metadata->data_tail;
+	mm->stalled = 0;
 
 	return (0);
 }
@@ -667,6 +670,9 @@ perf_mmap_read(struct perf_mmap *mm, u8 *wrapped_event_buf)
 	u16				 evsize;
 	ssize_t				 leftcont;	/* contiguous size left */
 
+	if (unlikely(mm->stalled))
+		return (NULL);
+
 	data_head = perf_mmap_load_head(mm->metadata);
 	diff = data_head - mm->data_tmp_tail;
 	evh = (struct perf_event_header *)
@@ -679,16 +685,22 @@ perf_mmap_read(struct perf_mmap *mm, u8 *wrapped_event_buf)
 	 * Snapshot the size, the ring is a shared mapping and the checks below
 	 * are useless against a value that changes under our feet.
 	 */
-	evsize = evh->size;
+	evsize = __atomic_load_n(&evh->size, __ATOMIC_RELAXED);
 	/*
 	 * A record smaller than its header is corruption, consuming it would
-	 * spin on the same offset forever, treat the ring as empty and let it
-	 * stall instead. Nothing here can overflow the copies below: evsize
-	 * caps at UINT16_MAX, wrapped_event_buf holds UINT16_MAX + 1 (see
+	 * spin on the same offset forever, warn once and stall the ring
+	 * instead. Nothing here can overflow the copies below: evsize caps at
+	 * UINT16_MAX, wrapped_event_buf holds PERF_RECORD_MAX_SIZE (see
 	 * kprobe_queue_open()) and the ring is at least as big (see the static
 	 * assert at PERF_MMAP_PAGES), even against a corrupt data_head.
 	 */
-	if (evsize < sizeof(*evh) || diff < evsize)
+	if (unlikely(evsize < sizeof(*evh))) {
+		mm->stalled = 1;
+		qwarnx("perf record is corrupt: %d bytes", evsize);
+		return (NULL);
+	}
+	/* Do we have the full record */
+	if (diff < evsize)
 		return (NULL);
 	/* How much contiguous space there is left */
 	leftcont = mm->data_size - (mm->data_tmp_tail & mm->data_mask);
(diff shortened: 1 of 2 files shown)


nicholasberlin:
CI failures root-caused and fixed in e15a86b: kernels older than 5.16 cap perf tracepoint records at `PERF_MAX_TRACE_SIZE` (2048) and silently drop anything bigger, so the PATH_MAX exec record the test relies on never arrives (exe stays inherited from fork and the strcmp assert fired). Backports make version checks unreliable — Ubuntu's 5.4 drops the record while RHEL's 4.18.0-553 delivers it — so the test now probes the behavior: a mid-sized path below the cap is asserted strictly on every kernel, and the full PATH_MAX check applies only when the oversized record is actually delivered, skipping otherwise (a *partial* copy of the long path still fails, since that would be quark corrupting the record rather than the kernel dropping it).

Note this also means the fatal errx being removed here was unreachable on those older kernels — no record over 2048 bytes could be delivered — so skipping the strict check there matches reality.

Verified locally: 5.4.0-150-generic (skips), 4.18.0-553.el8_10 (strict), 6.8.0-65-generic (strict); full suite green on all three.

kprobe_queue.c
```diff
@@ -651,42 +659,55 @@ perf_mmap_update_tail(struct perf_event_mmap_page *metadata, uint64_t tail)
...
-	/* Do we have at least one complete event */
-	if (diff < (int)sizeof(*evh) || diff < evh->size)
+	/* Do we have at least one complete header */
+	if (diff < (int)sizeof(*evh))
+		return (NULL);
+	/*
+	 * Snapshot the size, the ring is a shared mapping and the checks below
+	 * are useless against a value that changes under our feet.
+	 */
+	evsize = evh->size;
```
christos68k: We should make this an atomic read as described [here](https://github.com/libbpf/libbpf/blob/f7081a6baf3f54949aacb8c2fc11bb30783b83e9/src/libbpf.h#L1750-L1751):

(Alternatively, add a `static inline perf_mmap_load_evsize` function before `perf_mmap_read`)

```suggestion
	evsize = __atomic_load_n(&evh->size, __ATOMIC_RELAXED);
```

kprobe_queue.c
```diff
@@ -27,6 +27,13 @@
 #include "quark.h"
 
 #define PERF_MMAP_PAGES		16		/* Must be power of 2 */
+/*
+ * A perf record caps at UINT16_MAX (perf_event_header.size). The ring must be
+ * able to hold at least one such record, or perf_mmap_read() could chase a
+ * corrupt data_head past the end of the ring. getpagesize() is at least 4096.
+ */
+static_assert(PERF_MMAP_PAGES * 4096 >= UINT16_MAX + 1,
```
christos68k: We can turn this into a constant as it's used in multiple places:

```suggestion
#define PERF_RECORD_MAX_SIZE (UINT16_MAX + 1)

static_assert(PERF_MMAP_PAGES * 4096 >= PERF_RECORD_MAX_SIZE,
```

kprobe_queue.c
```diff
@@ -1384,6 +1405,18 @@ kprobe_queue_open(struct quark_queue *qq)
 	kqq->qid = qid;
 	qq->queue_be = kqq;
 
+	/*
+	 * Scratch space for perf_mmap_read() to linearize a record that wraps
+	 * the ring, shared by all rings since records are consumed one at a
+	 * time. A record caps at UINT16_MAX (perf_event_header.size is u16),
+	 * so any record fits.
+	 */
+	kqq->wrapped_event_buf = malloc(UINT16_MAX + 1);
```
christos68k:
```suggestion
	kqq->wrapped_event_buf = malloc(PERF_RECORD_MAX_SIZE);
```

kprobe_queue.c
```diff
@@ -651,42 +659,55 @@ perf_mmap_update_tail(struct perf_event_mmap_page *metadata, uint64_t tail)
...
+	/*
+	 * A record smaller than its header is corruption, consuming it would
+	 * spin on the same offset forever, treat the ring as empty and let it
+	 * stall instead. Nothing here can overflow the copies below: evsize
+	 * caps at UINT16_MAX, wrapped_event_buf holds UINT16_MAX + 1 (see
+	 * kprobe_queue_open()) and the ring is at least as big (see the static
+	 * assert at PERF_MMAP_PAGES), even against a corrupt data_head.
+	 */
+	if (evsize < sizeof(*evh) || diff < evsize)
 		return (NULL);
```
christos68k: We can split these checks into two branches to enable:

1. Keeping the `unlikely` for the unlikely branch
2. Calling `qwarnx` if a stall takes place (otherwise we're completely blind)
3. We can set `mm->stalled = 1` and then fast-exit from `perf_mmap_read`


```suggestion
	if (unlikely(evsize < sizeof(*evh))) {
		mm->stalled = 1;
		qwarnx("perf record is corrupt: %d bytes", evsize);
		return (NULL);
	}
	
	if (diff < evsize)
		return (NULL);
```

kprobe_queue.c
```diff
@@ -651,42 +659,55 @@ perf_mmap_update_tail(struct perf_event_mmap_page *metadata, uint64_t tail)
 }
 
 static struct perf_event *
-perf_mmap_read(struct perf_mmap *mm)
+perf_mmap_read(struct perf_mmap *mm, u8 *wrapped_event_buf)
 {
```
christos68k, line 666:
```suggestion
{
	if (unlikely(mm->stalled))
		return (NULL);
		
```

kprobe_queue.c
nicholasberlin: done [28428e0](https://github.com/elastic/quark/pull/385/commits/28428e043d0ffd0647589482866d42190b191a98)

kprobe_queue.c
nicholasberlin: [28428e0](https://github.com/elastic/quark/pull/385/commits/28428e043d0ffd0647589482866d42190b191a98)

kprobe_queue.c
nicholasberlin: [28428e0](https://github.com/elastic/quark/pull/385/commits/28428e043d0ffd0647589482866d42190b191a98)

kprobe_queue.c
nicholasberlin: [28428e0](https://github.com/elastic/quark/pull/385/commits/28428e043d0ffd0647589482866d42190b191a98)

kprobe_queue.c
nicholasberlin: [28428e0](https://github.com/elastic/quark/pull/385/commits/28428e043d0ffd0647589482866d42190b191a98)

nicholasberlin:
@christos68k thanks for the review.

`qwarnx` does not actually bubble up to quark library users. What do you think about adding something about stalled queues in `quark_queue_get_stats`? another PR or fold into this one?

christos68k (approved):
LGTM

christos68k:
> @christos68k thanks for the review.
> 
> `qwarnx` does not actually bubble up to quark library users. What do you think about adding something about stalled queues in `quark_queue_get_stats`? another PR or fold into this one?

A new PR is fine, I can do this today, will help me look at more quark code.

PR #386, opened 2026-08-13
go: expose rulesets, passwd/group lookups and sock established

nicholasberlin:
Closes the three gaps Go consumers are currently working around:

- **Rule DSL**: `QueueAttr.RuleText` installs a ruleset on the queue. It's parsed with `quark_ruleset_parse()` inside `OpenQueue()` — before any privileged operation, so parse errors surface without root — via an fmemopen shim borrowed from quark-test's `ruleset_from_string1()`. The C ruleset is owned by the `Queue` and released on `Close()` (the queue only borrows the pointer). Processes matched by poison rules carry the tag in the new `Process.PoisonTag` (set unconditionally, zero = no match, same semantics as `quark_process.poison_tag`). Until now the DSL — including `file.exec_change` — was unusable from Go.
- **User/group names**: `PasswdLookup()` and `GroupLookup()` mirror quark_passwd_lookup(3) and quark_group_lookup(3). The otel quark receiver currently skips `user.name`/`group.name` in its ECS output with a comment noting the binding doesn't exist.
- **`QUARK_EV_SOCK_CONN_ESTABLISHED`**: the only event bit missing from the bindings (`..._CLOSED` was there); the otel receiver hand-rolls `uint64(1) << 5` today.

Tests: `TestRuleText` covers parse success/failure without root; `TestQuark` gains `PasswdGroupLookup` (uid/gid 0 resolve to root, unknown ids miss) and `RulePoison`, a Go port of `t_rule_poison` (poison children by ppid, pass only poisoned, drop the rest — every surviving event must carry the tag). RulePoison uses `/bin/true` rather than `/bin/echo` so it also runs in the initramfs VM.

Verified: full Go suite passes as root on real hardware (`sudo ./quark-go-test -test.v`, all 7 TestQuark subtests + TestRuleText); the new subtests also pass under krun with an initramfs augmented with glibc and /etc/passwd. go vet and gofmt clean.

Next tier of exposure candidates (entity_id, change_mask, taints, id_change, packet payload) deliberately left out to keep this reviewable; container metadata is already in flight in #373/#380–383.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -143,3 +143,8 @@ Changes from 0.7 to 0.8
     another backend themselves. QQ_ALL_BACKENDS was removed and
     breaks at compile time. quark_queue_default_attr() now selects
     only QQ_EBPF.
+  o The Go bindings (go/quark) learned PasswdLookup() and
+    GroupLookup(), mirroring quark_passwd_lookup(3) and
+    quark_group_lookup(3), QueueAttr.RuleText for installing a
+    ruleset on the queue, Process.PoisonTag and the
+    QUARK_EV_SOCK_CONN_ESTABLISHED constant.
--- a/go/quark/quark.go
+++ b/go/quark/quark.go
@@ -9,7 +9,9 @@ package quark
 #cgo CFLAGS: -I${SRCDIR}/../../
 #cgo LDFLAGS: -Wl,--wrap=fmemopen ${SRCDIR}/../../libquark_big.a
 
+#include <errno.h>
 #include <stdlib.h>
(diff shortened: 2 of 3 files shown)


nicholasberlin:
Superseded by single-concern PRs, one per concern:

- #390 — the QUARK_EV_SOCK_CONN_ESTABLISHED constant
- #391 — PasswdLookup/GroupLookup
- #392 — rulesets (QueueAttr.RuleText, Process.PoisonTag)

The code in those PRs is identical to this one.

PR #388, opened 2026-08-14
Document the boottime calls

nicholasberlin:
Split out of #372; this carries only the documentation.

Manpages for quark_update_boottime(3), quark_get_boottime(3) and quark_time_to_wallclock(3), listed in quark(7), with the HTML docs and README.md regenerated. The pages document which times are expressed since boot, that btime comes from /proc/stat and moves on clock steps, and two common ways of detecting a step (polling the CLOCK_REALTIME - CLOCK_BOOTTIME delta, or a timerfd armed with TFD_TIMER_CANCEL_ON_SET). Each page carries a HISTORY section noting the calls appeared in quark 0.8, when all times handed out by quark switched from epoch to since-boot, so readers on older versions aren't misled.

mandoc -Tlint passes on the new pages.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -335,7 +335,12 @@ These scripts require KVM access and therefore must be executed on a host kernel
   <li>No threads are created, the library is driven solely through
       <a class="Xr" href="quark_queue_get_event.3.html">quark_queue_get_event(3)</a>.</li>
   <li>Access to a <var class="Vt">quark_queue</var> must be synchronized by the
-      user in the case of multithreading.</li>
+      user in the case of multithreading. The boottime epoch is the exception:
+      <a class="Xr" href="quark_update_boottime.3.html">quark_update_boottime(3)</a>,
+      <a class="Xr" href="quark_get_boottime.3.html">quark_get_boottime(3)</a>
+      and
+      <a class="Xr" href="quark_time_to_wallclock.3.html">quark_time_to_wallclock(3)</a>
+      may be called from any thread without synchronization.</li>
 </ul>
 </section>
 <section class="Sh">
@@ -463,7 +468,7 @@ main(void)
 </div>
 <table class="foot">
   <tr>
-    <td class="foot-date">August 3, 2026</td>
+    <td class="foot-date">August 14, 2026</td>
     <td class="foot-os">Linux</td>
   </tr>
 </table>
--- a/docs/index.html
+++ b/docs/index.html
@@ -412,6 +412,12 @@ main(void)
   <dd>block for an unspecified amount of time.</dd>
   <dt><a class="Xr" href="quark_queue_get_stats.3.html">quark_queue_get_stats(3)</a></dt>
   <dd>basic queue statistics.</dd>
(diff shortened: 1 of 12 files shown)


nicholasberlin:
Pushed 4d96cf0 documenting that `quark_update_boottime()`, `quark_get_boottime()` and `quark_time_to_wallclock()` are safe to call from any thread without synchronization, as an exception to the quark(7) queue-synchronization convention.

**Merge order:** that statement is only true once #393 merges — it routes all access to the process-wide `quark.boottime`/`quark.hz` through relaxed `__atomic` helpers (codegen-identical on x86-64/aarch64, builds on the centos7 gcc 4.8.5). Today the epoch is a plain u64, so concurrent `UpdateBoottime`/`TimeToWallclock` — as auditbeat does across two independent consumers in one process — is a data race. #393 needs to land before this PR.

docs/index.html
```diff
@@ -335,7 +335,12 @@ <h1 class="Sh" id="CONVENTIONS"><a class="permalink" href="#CONVENTIONS">CONVENT
   <li>No threads are created, the library is driven solely through
       <a class="Xr" href="quark_queue_get_event.3.html">quark_queue_get_event(3)</a>.</li>
   <li>Access to a <var class="Vt">quark_queue</var> must be synchronized by the
-      user in the case of multithreading.</li>
+      user in the case of multithreading. The boottime epoch is the exception:
+      <a class="Xr" href="quark_update_boottime.3.html">quark_update_boottime(3)</a>,
+      <a class="Xr" href="quark_get_boottime.3.html">quark_get_boottime(3)</a>
+      and
+      <a class="Xr" href="quark_time_to_wallclock.3.html">quark_time_to_wallclock(3)</a>
+      may be called from any thread without synchronization.</li>
```
biscout42, line 343: > without synchronization

is it because of https://github.com/elastic/quark/pull/393 ?

quark_time_to_wallclock.3
```diff
@@ -0,0 +1,67 @@
...
+.Em established_time
+and
+.Em close_time
+of
+.Vt quark_socket .
+These fields are expressed as nanoseconds since boot since quark 0.8,
+in earlier versions they were nanoseconds since the Unix epoch and
+needed no translation.
+.Pp
+Times since boot are immune to system clock changes, wallclock times
```
biscout42, line 38: nit: about wording, how about to mention `clock skew` to make explicit what is the problem we are trying to solve?

biscout42 (approved):
LGTM; I would appreciate a second review from a person with more experience in the times we collect for quark events to verify the docs.

docs/index.html
nicholasberlin: yes, this PR will merge after that PR

quark_time_to_wallclock.3
nicholasberlin: "Step" is deliberate: skew gets corrected by slewing, which adjusts the wallclock and the boot clock at the same rate — the epoch is their difference, so it never moves and the translation is unaffected. Only a step moves it.

Added a bit in commit e16f3ee to clarify.

Thanks.

PR #390, opened 2026-08-14
go: expose QUARK_EV_SOCK_CONN_ESTABLISHED

nicholasberlin:
Split out of #386; this carries only the event constant.

QUARK_EV_SOCK_CONN_ESTABLISHED was the only event bit missing from the bindings (`..._CLOSED` was there); the otel receiver hand-rolls `uint64(1) << 5` today.

Note: the go/quark PRs split out of #372/#386 each append a CHANGES bullet, so whichever merges later needs a trivial rebase.


biscout42 (approved):
For my understanding: why don't we expose `QUARK_EV_ID_CHANGE` and `QUARK_EV_GETPID`?

nicholasberlin:
> For my understanding: why don't we expose QUARK_EV_ID_CHANGE and QUARK_EV_GETPID?

Deliberate scope call for the 0.8 series (https://github.com/elastic/quark/pull/388, https://github.com/elastic/quark/pull/391, https://github.com/elastic/quark/pull/392, and https://github.com/elastic/quark/pull/393): these PRs only expose what the consumers being migrated right now (auditbeat, the otel receiver, odinq) actually use.

`QUARK_EV_SOCK_CONN_ESTABLISHED` is in a different bucket than `QUARK_EV_ID_CHANGE` and `QUARK_EV_GETPID`. The go bindings already expose `QQ_SOCK_CONN`, the Socket payload and `QUARK_EV_SOCK_CONN_CLOSED`, so established events were already being delivered to Go users. There was just no named bit to test Events against. This PR completes an existing surface rather than opening a new one.


`ID_CHANGE` would open a new surface that only makes sense as a set. `QUARK_EV_ID_CHANGE` alone is a bit you can see with no data to act on — it needs the `id_change` field plumbed through `Process`/`Event` plus the `QUARK_ID_CHANGE_*` mask. So, a separate PR.

`QUARK_EV_GETPID` is just a development/testing flag, and should not be exposed.

`ID_CHANGE` is on the follow-up list for when a Go consumer needs them, same bucket as `entity_id`, the file `change_mask`, kernel module `taints`, and the packet payload. Or we could just add them later and not wait for a consumer.

PR #392, opened 2026-08-14
go: expose rulesets

nicholasberlin:
Split out of #386; this carries only the rule DSL exposure.

`QueueAttr.RuleText` installs a ruleset on the queue. It's parsed with `quark_ruleset_parse()` inside `OpenQueue()` — before any privileged operation, so parse errors surface without root — via an fmemopen shim borrowed from quark-test's `ruleset_from_string1()`. The C ruleset is owned by the `Queue` and released on `Close()` (the queue only borrows the pointer). Processes matched by poison rules carry the tag in the new `Process.PoisonTag` (set unconditionally, zero = no match, same semantics as `quark_process.poison_tag`). Until now the DSL — including `file.exec_change` — was unusable from Go.

Tests: `TestRuleText` covers parse success/failure without root; `TestQuark` gains `RulePoison`, a Go port of `t_rule_poison` (poison children by ppid, pass only poisoned, drop the rest — every surviving event must carry the tag). RulePoison uses `/bin/true` rather than `/bin/echo` so it also runs in the initramfs VM.

Code identical to #386, where the full Go suite passed as root on real hardware and under krun; gofmt, go vet and a test-binary link re-checked here.

Note: the go/quark PRs split out of #372/#386 each append a CHANGES bullet, so whichever merges later needs a trivial rebase.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -148,3 +148,6 @@ Changes from 0.7 to 0.8
     and TimeToWallclock(), mirroring the C calls.
   o The Go bindings (go/quark) learned the
     QUARK_EV_SOCK_CONN_ESTABLISHED constant.
+  o The Go bindings (go/quark) learned QueueAttr.RuleText for
+    installing a ruleset on the queue. Processes matched by poison
+    rules carry the tag in Process.PoisonTag.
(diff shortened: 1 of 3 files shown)


mjwolf (approved):
LGTM

PR #395, opened 2026-08-20
Link container data regardless of kube talker

biscout42:
As we link container data from an external process there is no need to check kube talker. 

That is a bit complex, because adds parsing cgroup to _all_ events. I have added guards to reduce the performance footprint for non-k8s use case. It should behave like this: 
  - An empty metadata tree causes no cgroup parse.
  - A non-container cgroup is parsed only once.
  - A container ID is built and allocated only once.
  - A failed metadata lookup still occurs for each event while the tree is nonempty.
  - A cgroup change clears the cached result.
  - Container metadata that arrives later can still match the cached ID.

Implications: 
  - Initial metadata: all existing processes are linked in a full pass during a queue open.
  - Later asynchronous metadata: no immediate full pass; processes are linked opportunistically when they next produce an event.
  - A process that produces no subsequent event **can remain unlinked** until another linking pass or relevant event occurs.

Imho the last point is acceptable because no events meaning no messages to deliver from a queue. 

TODO: 

- [x] Backend-less test fixture. #395 adds a shared quark_queue_init_trees helper plus its own stub queue ops in the C tests, but does not consolidate the two fixtures from #383 or the copy in the Go cgo preamble. There are now four copies instead of three.
- [x] Explore alternative for dropping the `quarktest` tag. In this PR libquark exports quark_queue_init_trees in the header's "exported for testing only" section. With a one-line null guard on queue_ops in quark_queue_populate, matching the guard already in quark_queue_close, the stub ops become unnecessary. The Go helper then shrinks to a few lines that call calloc, call the init function, and set two fields, with no C preamble at all. It could sit unexported at the bottom of quark.go. That is test-only code in a production file, but it mirrors the precedent libquark already sets in its header. Precisely the same change also collapses the three hand-rolled C and Go fixtures, which is the fixture-duplication finding two items down the list.

### DIFF
--- a/quark.c
+++ b/quark.c
@@ -4919,7 +4919,7 @@ raw_event_sock(struct quark_queue *qq, struct raw_event *raw)
 	qsk->pid_last_use = raw->pid;
 
 	qev->socket = qsk;
-	qev->process = quark_process_lookup(qq, qsk->pid_origin);
+	qev->process = process_cache_get(qq, qsk->pid_origin, 0);
 
 	return (qev);
 }
@@ -4938,7 +4938,7 @@ raw_event_packet(struct quark_queue *qq, struct raw_event *raw)
 	qev = &qq->event_storage;
 
 	qev->events = QUARK_EV_PACKET;
-	qev->process = quark_process_lookup(qq, raw->pid);
+	qev->process = process_cache_get(qq, raw->pid, 0);
 
 	/* Steal the packet */
 	qev->packet = raw->packet.quark_packet;
@@ -4963,7 +4963,7 @@ raw_event_file(struct quark_queue *qq, struct raw_event *raw)
 	qev = &qq->event_storage;
 
 	qev->events = QUARK_EV_FILE;
-	qev->process = quark_process_lookup(qq, raw->pid);
+	qev->process = process_cache_get(qq, raw->pid, 0);
 
 	/*
 	 * File aggregation is basically joining op_mask and then using the last
@@ -4994,7 +4994,7 @@ raw_event_ptrace(struct quark_queue *qq, struct raw_event *raw)
 	qev = &qq->event_storage;
 
 	qev->events = QUARK_EV_PTRACE;
-	qev->process = quark_process_lookup(qq, raw->pid);
+	qev->process = process_cache_get(qq, raw->pid, 0);
 	qev->ptrace = raw->ptrace.quark_ptrace;
 
 	return (qev);
@@ -5008,7 +5008,7 @@ raw_event_module_load(struct quark_queue *qq, struct raw_event *raw)
 	qev = &qq->event_storage;
 
 	qev->events = QUARK_EV_MODULE_LOAD;
-	qev->process = quark_process_lookup(qq, raw->pid);
+	qev->process = process_cache_get(qq, raw->pid, 0);
 	/* Steal it */
 	qev->module_load = raw->module_load.quark_module_load;
 	raw->module_load.quark_module_load = NULL;
@@ -5024,7 +5024,7 @@ raw_event_shm(struct quark_queue *qq, struct raw_event *raw)
 	qev = &qq->event_storage;
 
 	qev->events = QUARK_EV_SHM;
-	qev->process = quark_process_lookup(qq, raw->pid);
+	qev->process = process_cache_get(qq, raw->pid, 0);
 	/* Steal it */
 	qev->shm = raw->shm.quark_shm;
 	raw->shm.quark_shm = NULL;
@@ -5047,7 +5047,7 @@ raw_event_tty(struct quark_queue *qq, struct raw_event *raw)
 	qev = &qq->event_storage;
 
 	qev->events = QUARK_EV_TTY;
-	qev->process = quark_process_lookup(qq, raw->pid);
+	qev->process = process_cache_get(qq, raw->pid, 0);
 
 	/* Link raw (our head) to the first chunk in tha agg queue */
 	next = TAILQ_FIRST(&raw->agg_queue);
--- a/quark.c
+++ b/quark.c
@@ -1582,7 +1582,7 @@ parse_container_cgroup(const char *cgroup, char *container_id, size_t container_
 }
 
 static void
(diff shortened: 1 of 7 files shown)


quark.c
```diff
@@ -4466,15 +4467,11 @@ quark_queue_open(struct quark_queue *qq, struct quark_queue_attr *qa)
...
+	 * At this point, existing processes and container metadata have been
+	 * loaded. Now it is time to correlate them.
 	 */
-	if (qq->qkube != NULL) {
-		struct quark_process	*qp;
-
-		RB_FOREACH(qp, process_by_pid, &qq->process_by_pid) {
-			link_kube_data(qq, qp);
-		}
+	RB_FOREACH(qp, process_by_pid, &qq->process_by_pid) {
```
Tacklebox, line 4547: Hmm, I have been testing with this loop basically the same but in the quark_container_create function since now the kube/container data is coming from outside of quark after we open the queue. If there's a ton of existing processes it's very inefficient but we need to do it there if we are to support the async k8s socket fallback. Also, I'm not sure this comment is true anymore.

biscout42: As it happens only during the startup imho performance penalty is acceptable. How would you suggest to fix it? My understanding that w/o this call current changes in the stack will not attach container metadata to existing processes. Is that correct understanding?

quark.c
```diff
@@ -5369,10 +5366,8 @@ quark_queue_get_event1(struct quark_queue *qq)
 	if (qev == NULL)
 		return (NULL);
 
-	/* Try to correlate kubernetes metadata */
-	if (qq->qkube != NULL)
-		link_kube_data(qq,
-		    (struct quark_process *)qev->process);
+	/* Try to correlate container metadata. */
+	link_container_data(qq, (struct quark_process *)qev->process);
```
biscout42, line 5444: @Tacklebox , I'm wondering how did you test your changes? Did you have kube talker running as well?

quark.c
```diff
@@ -1581,21 +1582,81 @@ parse_container_cgroup(const char *cgroup, char *container_id, size_t container_
...
+	/* Do not parse the cgroup if no container metadata exists. */
+	if (RB_EMPTY(&qq->container_by_id))
 		return;
-	if (parse_container_cgroup(qp->cgroup, cid, sizeof(cid)) == -1)
+
+	container_id = process_container_id(qp);
+	if (container_id == NULL)
 		return;
-	if ((container = container_lookup(qq, cid)) == NULL)
+	if ((container = container_lookup(qq, (char *)container_id)) == NULL)
```
christos68k: Instead of the hard cast, we can update `container_lookup` to take `const char *` as an argument which also lets the developer know that it doesn't mutate the string.

quark.c
```diff
@@ -1581,21 +1582,81 @@ parse_container_cgroup(const char *cgroup, char *container_id, size_t container_
...
+		return (NULL);
+
+	if (parse_container_cgroup(qp->cgroup, cid, sizeof(cid)) == -1) {
+		qp->container_id_parsed = 1;
+		return (NULL);
+	}
+
+	qp->container_id = strdup(cid);
+	if (qp->container_id == NULL)
+		return (NULL);
```
christos68k, line 1608: One issue here is that `strdup` will return `NULL` if it can't allocate (e.g. cgroup memory limits for the container are reached). Since we're not setting `qp->container_id_parsed = 1` here (as opposed to the `parse_container_cgroup` failure) we're saying that we can keep trying to parse the cgroup in future calls. 

One alternative is to set `qp->container_id_parsed = 1` and treat the allocation failure as a general failure. However we can't distinguish between "this is not a container" and "this is a container but we failed during extraction".

quark.c
christos68k: If the K8s data is coming from outside of quark after the queue is opened, isn't this logic always a NOP if there's no `qq->qkube`?

1. `quark_queue_open` calls `RB_FOREACH` (outside of the `qa->kubefd != -1` branch)
2. `RB_FOREACH` calls `link_container_data`
3. `link_container_data` calls [RB_EMPTY(&qq->container_by_id)](https://github.com/elastic/quark/blob/8eeb274190cdab6846d46e93085359c1945754d7/quark.c#L1653)

However, `quark_queue_open` also calls [RB_INIT(&qq->container_by_id)](https://github.com/elastic/quark/blob/8eeb274190cdab6846d46e93085359c1945754d7/quark.c#L4395-L4395) which means the tree is empty at function entry and stays empty if there's no `qq->qkube`.

biscout42: @christos68k , I leave it here only because we do not remove (yet) kube talker. My understanding we need this `RB_FOREACH` only in case we run quark together with a kube talker. Please, let me know if I missed something.

quark.c
biscout42: Yeah, idea here was to re-try, in case of a failure. Do you see a better option? Please, share. FYI: we have quite high limits by default, here https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-managed/elastic-agent-managed-daemonset.yaml#L92

quark.c
biscout42: Thanks, applied.

quark.c
christos68k: Maybe add a comment to clarify that `strdup` returning `NULL` is taken into account e.g.

```suggestion
		// If strdup returns NULL, don't set qp->container_id_parsed to enable retries
		return (NULL);
```

quark.c
christos68k: I mean we should keep the `RB_FOREACH(qp, process_by_pid, &qq->process_by_pid)` inside the `if (qq->qkube != NULL) {` guard, because if `qq->qkube == NULL` it doesn't have any effect, it will keep running into `RB_EMPTY` and wasting CPU cycles.

biscout42: That is actually very good catch, thank you. 

IMHO we need to make a design decision here: do we allow a quark user to supply container metadata before `quark_queue_open` (IIUC kube talker was able to do so due-to a pipe fd)? @Tacklebox / @christos68k , WDYT?

biscout42: @christos68k , done.

PR #396, opened 2026-08-20
feat: add container / host event scope to ruleset

biscout42:
The change allows to filter host events vs container events.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -126,6 +126,9 @@ Changes from 0.6 to 0.7
 
 Changes from 0.7 to 0.8
 ~~~~~~~~~~~~~~~~~~~~~~~
+  o The rule DSL learned the event.scope field. Its values are container
+    and host. The field is not supported by the nova backend, where it
+    fails with ENOTSUP.
   o BREAKING, SILENTLY: quark_process.proc_time_boot,
     quark_process.exit_time_event and quark_event.time are now
     expressed in nanoseconds since boot instead of nanoseconds since
--- a/quark-mon.8
+++ b/quark-mon.8
@@ -125,6 +125,14 @@ Read rules from
 use
 .Em -
 for stdin.
+The
+.Dq event.scope container
+field matches events from a process with a container ID.
+The
+.Dq event.scope host
+field matches events from a process without a container ID.
+An event without a process matches neither scope.
+This field is not supported by the NOVA backend.
 .It Fl T
 Enable ptrace event tracing.
 .It Fl t
--- a/quark.c
+++ b/quark.c
@@ -5144,6 +5144,7 @@ quark_rule_field_match(struct quark_rule *rule, struct quark_rule_field *field,
     struct quark_event *qev)
 {
 	const struct quark_process	*qp;
+	const char			*container_id;
 
 	qp = qev->process;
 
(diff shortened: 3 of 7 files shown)


Tacklebox (commented):
I'm not sure it makes sense to write rules based on container.id outside of the special case of LXC, since the ids won't be known at the time when the policy is written since they are generated at runtime right? I don't see any technical issues with the PR though.

christos68k (dismissed):

biscout42:
I'll update this logic when https://github.com/elastic/quark/pull/395 is done.

quark.c
```diff
@@ -5182,6 +5183,15 @@ quark_rule_field_match(struct quark_rule *rule, struct quark_rule_field *field,
 			    (qev->file->change_mask & QUARK_FILE_CH_PERMS)) &&
 			    (qev->file->mode & (S_IXUSR | S_IXGRP | S_IXOTH)));
 		break;
+	case QUARK_RF_EVENT_SCOPE:
+		if (qp == NULL)
+			break;
+		container_id = process_container_id((struct quark_process *)qp);
```
biscout42, line 5189: This is key point: `process_container_id` will treat `container_id_parsed` and also will use a cached value, if we already parsed container_id.

biscout42:
@Tacklebox , the use case here is to separate container events for k8s vs host events.

CHANGES
```diff
@@ -192,3 +192,11 @@ Changes from 0.8 to 0.9
...
     quark_shm memfd_create_flags and shmget_{key,size,shmflg}, which were
     read from the wrong offsets on that kernel.
+  o The rule DSL learned the event.scope field. Its values are container
+    and host. A container event comes from a process whose cgroup names
+    a container, a host event from a process whose cgroup does not. A
+    process with an unknown cgroup matches neither scope. The kprobe
+    backend does not report the cgroup of new processes, so on kprobe
+    only processes seen at start-up can match a scope. Only the last
+    cgroup component is parsed, nested cgroups match host. The nova
+    backend rejects the field with ENOTSUP.
```
christos68k, line 202: Suggest rewording this for clarity as I got a little confused when I first read it:

```suggestion
  o The rule DSL learned the event.scope field, which selects events by
    where the originating process runs: "container" matches processes
    running inside a container, "host" matches processes running directly
    on the host. A process is inside a container when the last component
    of its cgroup path is a systemd scope unit created by a container
    runtime, like docker-<id>.scope. Any other cgroup path is treated as
    host, including nested cgroups inside a container and the flat layout
    of the cgroupfs driver. Events without a process, or from a process
    whose cgroup is unknown, match neither value. The KPROBE backend only
    learns the cgroup of processes that existed when quark started, so events
    from processes created afterwards never match either value. The NOVA
    backend rejects the field with ENOTSUP.
```

quark-mon.8
```diff
@@ -125,6 +125,21 @@ Read rules from
...
+field matches events from a process whose cgroup does not name a container.
+An event without a process, or from a process with an unknown cgroup,
+matches neither scope.
+The kprobe backend does not report the cgroup of new processes, so on
+kprobe only processes that existed at start-up can match a scope.
+Only a cgroup whose last component is a container scope, like
+.Pa docker-<id>.scope ,
+names a container.
+Nested cgroups and the cgroupfs driver layout match the host scope.
+This field is not supported by the NOVA backend.
```
christos68k, line 142: Same here, reworded for clarity:

```suggestion
The
.Dq event.scope
field matches events by where the originating process runs.
.Dq event.scope container
matches events from processes running inside a container and
.Dq event.scope host
matches events from processes running directly on the host.
A process is considered to be inside a container when the last component
of its cgroup path is a systemd scope unit created by a container runtime,
like
.Pa docker-<id>.scope .
Any other cgroup path is treated as host.
This includes a nested cgroup inside a container, like
.Pa docker-<id>.scope/init.scope ,
and the flat layout of the cgroupfs driver, like
.Pa /docker/<id> .
Events without a process, and events from a process whose cgroup is
unknown, match neither value.
The kprobe backend only learns the cgroup of processes that existed when
.Nm
started, so events from processes created afterwards never match either
value.
This field is not supported by the NOVA backend.
```

christos68k (approved):
Left some documentation nits for clarity.

LGTM

christos68k (dismissed):

CHANGES
biscout42: after resolving conflicts while rebase this happened to be in https://github.com/elastic/quark/pull/430/changes

PR #397, opened 2026-08-24
feat: go-quark learns disable aggregation

biscout42:


go/quark/quark.go
```diff
@@ -577,6 +577,19 @@ func TimeToWallclock(timeSinceBoot uint64) uint64 {
 	return uint64(C.quark_time_to_wallclock(C.u64(timeSinceBoot)))
 }
 
+// DisableAggregation clears every entry in Quark's aggregation matrix.
+// It must be called after OpenQueue and before the queue is consumed.
+func (queue *Queue) DisableAggregation() error {
+	for parent := 0; parent < int(C.RAW_NUM_TYPES); parent++ {
+		for child := 0; child < int(C.RAW_NUM_TYPES); child++ {
+			if C.quark_queue_set_agg_matrix(queue.quarkQueue, C.int(parent), C.int(child), nil) != 0 {
+				return wrapErrno(errors.New("clear aggregation matrix"))
```
christos68k: There's no `errno` here to wrap (the error is created by us, inline):


```suggestion
				return errors.New("clear aggregation matrix")
```

christos68k (approved):

go/quark/quark.go
biscout42: good catch, thanks. It actually should be here. could you check the next commit?

christos68k: That looks fine. We can also skip `errno` checking and processing entirely in Go, as its value is fixed by the quark C code (`EINVAL`) and we're not doing anything with it in Go, but regardless, the current commit also works.

biscout42: Yeah, thanks. I followed current style. If you see how could we improve golang bindings, sure, lets do it. IMHO the bindings are small and we could iterate quite quick.

PR #399, opened 2026-08-26
Track perf ring stalls in quark_queue_get_stats

christos68k:
### Summary

Adds a counter for perf rings that are stalled due to corruption.

**TODO:**
- [x] Update manpage and html

Also see:
- #385
- https://github.com/elastic/quark/pull/385#issuecomment-5429783796

### DIFF
--- a/docs/quark_queue_get_stats.3.html
+++ b/docs/quark_queue_get_stats.3.html
@@ -41,6 +41,8 @@
 	u64	aggregations;
 	u64	non_aggregations;
 	u64	lost;
+	u64	garbage_collections;
+	u64	stalls;
 	int	backend;
 };</pre>
 </div>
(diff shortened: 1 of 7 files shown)


docs/quark_queue_get_stats.3.html
```diff
@@ -41,6 +41,8 @@ <h1 class="Sh" id="DESCRIPTION"><a class="permalink" href="#DESCRIPTION">DESCRIP
 	u64	aggregations;
 	u64	non_aggregations;
 	u64	lost;
+	u64	garbage_collections;
```
christos68k, line 44: This was missing from manpage/html so I added it here

quark.h
```diff
@@ -866,8 +866,9 @@ struct quark_queue_stats {
 	u64	non_aggregations;
 	u64	lost;
 	u64	garbage_collections;
-	int	backend;	/* active backend, QQ_EBPF or QQ_KPROBE */
-	/* TODO u64	peak_nodes; */
+	u64	stalls;     /* stalled perf rings due to corruption, only for QQ_KPROBE */
```
biscout42, line 869: cloud you pass it to [golang](https://github.com/elastic/quark/blob/main/go/quark/quark.go#L322)  bindings as well?

nicholasberlin: agreed

nicholasberlin (approved):
LGTM after adding the go-bindings

quark.h
christos68k: Done in e22740da5dc643ecf06b2b8374098c3a3045efa2

PR #367, opened 2026-07-17
Add mprotect syscall events

Aegrah:
## Summary

Collect successful executable-memory `mprotect()` calls as an eBPF-only Quark event.

`sys_enter_mprotect` records address, length, and protection when `PROT_EXEC` is requested. `sys_exit_mprotect` correlates the call, deletes its state, and emits only when the syscall succeeded. Emitted masks are X, RX, WX, and RWX; non-executable masks and failed calls are suppressed. No kprobe fallback is added.

Issue: https://github.com/elastic/endpoint-dev/issues/20408

## Workflow and proof

The contract workflow built and tested Quark, regenerated package/schema outputs, built Endpoint, ran focused unit/EAF tests, deployed through Fleet, and asserted Elasticsearch output.

- Preflight `20260722T124741Z-mprotect-preflight`: **PASS**
- Fleet acceptance `20260722T142745Z-mprotect` on Stack/Agent/Endpoint 9.4.2: **PASS**
- Elasticsearch semantic assertion: four X/RX/WX/RWX documents checked
- Realistic workstation qualification `20260722T103000Z-mprotect-realistic-smoke`: **PASS**

```text
3,708 Endpoint documents / 4,737 action values
exec: 523 (11.04%); mprotect: 6 (0.13%, 0.16% document coverage)
mprotect executables: /usr/bin/node 5; chromium headless_shell 1
masks: RWX 6 (100%)
```

## Companion PR

- endpoint-package: https://github.com/elastic/endpoint-package/pull/767

PR #400, opened 2026-08-27
Collect executable mprotect attempts at LSM hook

nicholasberlin:
## Summary

Stacked on #367.

- replace the paired `sys_enter_mprotect` / `sys_exit_mprotect` tracepoints and per-thread state with an fentry/kprobe pair on `security_file_mprotect`
- emit one event per VMA when the kernel-adjusted protection includes execute permission, covering `mprotect`, `pkey_mprotect`, compat callers, and `READ_IMPLIES_EXEC` handling through the common hook
- expose VMA bounds, normalized previous protection, requested and effective protection, and file-backed inode/device identity
- expose the richer event through the C and Go APIs and update event dumping and documentation
- restore `t_mprotect` to the RHEL 9.3 CI invocation now that it no longer depends on the affected syscall tracepoints

## Event semantics

An event means the kernel reached `security_file_mprotect` for this VMA with effective execute permission. The hook runs before the protection change is committed, so the event is an attempt and does not claim syscall success. The VMA bounds provide context and are not necessarily the exact caller-requested subrange.

## Verification

- `git diff --check`: pass
- `gofmt -d go/quark/quark.go`: no output
- Runtime/build tests not run: implementation was prepared on a macOS host, per request.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -394,7 +394,7 @@ steps:
 
   - label: "quark-test on rhel 9.3 (memfd+shmget broken)"
     key: test_rhel_9_3
-    command: "./.buildkite/runtest_distro.sh rhel 9.3 -x t_memfd -x t_shmget -x t_mprotect"
+    command: "./.buildkite/runtest_distro.sh rhel 9.3 -x t_memfd -x t_shmget"
     depends_on:
       - make_docker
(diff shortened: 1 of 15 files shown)


nicholasberlin:
Superseded by #404, which is now split into #409, #410, #411, #412 and #413.

PR #401, opened 2026-08-27
kprobe: translate MONOTONIC to BOOTTIME event timestamps

christos68k:
### Summary

The kprobe timestamps are currently always monotonic and do not include system suspend time (unlike the timestamps from the other backends, which are also monotonic but do include system suspend time). 

On kernels 4.1+ or kernels that have backported perf clockid, we can do better and have quark translate MONOTONIC to BOOTTIME internally. This matches the timestamps of the other backends. Roughly-monotonic timestamps are now dropped completely as we don't need to support any targets where perf clockid isn't available.

e4e6074a469b181819003835af68e83580fd3974 introduces a breaking API change with the drop of `QQ_MONOTONIC` which isn't needed any more.

@nicholasberlin noticed that we didn't have any RHEL7 kernel tests. This PR adds passing and a single failing test for RHEL7 kernels covering all three cases:

1. Pre-7.4 (negative test, checks for clean failure)
2. 7.4 (lowest supported RHEL7 kernel)
3. 7.9 (latest RHEL7 version)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -159,3 +159,11 @@ Changes from 0.7 to 0.8
   o The Go bindings (go/quark) learned Queue.DisableAggregation(),
     which clears every entry in the aggregation matrix. Call it after
     OpenQueue() and before the queue is consumed.
+
+Changes from 0.8 to 0.9
+~~~~~~~~~~~~~~~~~~~~~~~
+  o The KPROBE backend now translates event times from CLOCK_MONOTONIC
+    to CLOCK_BOOTTIME, matching the other backends. This means that produced
+    timestamps now include the total suspended time, same as the other backends.
+    QQ_MONOTONIC is now only set on kernels older than 4.1, where a perf
+    clock can not be selected.
--- a/CHANGES
+++ b/CHANGES
@@ -165,7 +165,9 @@ Changes from 0.8 to 0.9
   o The KPROBE backend now translates event times from CLOCK_MONOTONIC to
     CLOCK_BOOTTIME, matching the other backends. This means that produced
     timestamps now include the total suspended time, same as the other
-    backends.
+    backends. One exception: events stamped before a suspend but still
+    unread when the host resumes shift late by up to that suspend duration,
+    all other events are exact.
   o BREAKING: QQ_MONOTONIC was removed and event times are always
     CLOCK_BOOTTIME. This effectively drops support for kernels without perf
     clockid support (< 4.1 or RHEL before 7.4).
--- a/CHANGES
+++ b/CHANGES
@@ -162,8 +162,9 @@ Changes from 0.7 to 0.8
 
 Changes from 0.8 to 0.9
 ~~~~~~~~~~~~~~~~~~~~~~~
-  o The KPROBE backend now translates event times from CLOCK_MONOTONIC
-    to CLOCK_BOOTTIME, matching the other backends. This means that produced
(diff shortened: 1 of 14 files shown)


kprobe_queue.c
```diff
@@ -564,9 +570,33 @@ perf_sample_to_raw(struct quark_queue *qq, struct perf_record_sample *sample)
...
+static u64
+boot_offset(void)
+{
+	struct timespec	bt, mt;
+
+	/* Read monotonic first so the difference can not go negative on
+         * hosts that never suspended. */
+	if (clock_gettime(CLOCK_MONOTONIC, &mt) == -1 ||
+	    clock_gettime(CLOCK_BOOTTIME, &bt) == -1)
+		return (0);
```
christos68k: This error exit is here for clarity and can never realistically happen on any kernel that quark executes on. Also, `quark_queue_open` performs this test and aborts early before we reach this point.

kprobe_queue.c
```diff
@@ -564,9 +570,33 @@ perf_sample_to_raw(struct quark_queue *qq, struct perf_record_sample *sample)
...
+	/*
+	 * MAYBE TODO(ck): if the process gets preempted at exactly between the
```
christos68k: I added this as MAYBE TODO because it's probably not worth the extra lines of code. I can revisit if we think we gotta have it.

CHANGES
```diff
@@ -162,4 +162,8 @@ Changes from 0.7 to 0.8
 
 Changes from 0.8 to 0.9
 ~~~~~~~~~~~~~~~~~~~~~~~
-  o quark_queue_stats{} got a "stalls" member.
+  o The KPROBE backend now translates event times from CLOCK_MONOTONIC
+    to CLOCK_BOOTTIME, matching the other backends. This means that produced
+    timestamps now include the total suspended time, same as the other backends.
+    QQ_MONOTONIC is now only set on kernels older than 4.1, where a perf
```
christos68k: If we don't care about these kernels (and we probably don't? as `5.8` is an implicit requirement due to `bpf_ktime_get_boot_ns`) I can drop the `QQ_MONOTONIC` bits entirely.

**EDIT:** Went ahead and did it with e4e6074a469b181819003835af68e83580fd3974 as it simplifies code and API even more.

init.c
```diff
@@ -136,8 +136,9 @@ main(int argc, char *argv[])
 				errx(1, "couldn't mount tracefs or debugfs");
 			}
 		}
+		/* No cgroup2 on kernels older than 4.5 */
 		if (mount("cgroup2", "/sys/fs/cgroup", "cgroup2", 0, NULL) == -1)
-			err(1, "mount /sys/fs/cgroup");
+			warn("mount /sys/fs/cgroup");
```
christos68k: This was breaking the tests on pre-4.5 kernels since 2025. I guess we didn't notice because we didn't have any pre-4.5 kernels but now we do.

kprobe_queue.c
christos68k: Done in 13999059c2d06a6e62deaaeafab1fea3e149520a

init.c
nicholasberlin: perhaps test `errno` for `ENODEV`, so failures on new kernels will continue to fail hard.

christos68k: Done in 163b77dad3016afd0130e2f12608d0d001d2a7b1

nicholasberlin (approved):

PR #402, opened 2026-08-29
fix(ebpf): use ebpf_ringbuf_write in network and process probes

### LINKED ISSUE #288: Go over all probes and make sure they use ebpf_ringbuf_write
ebpf_ringbug_write is a wrapper that counts lost events in case the ring is full.
Some probes are still not using it, and should be converted.

amarkdotdev:
Several fixed-size network and process probe paths still called `bpf_ringbuf_reserve`/`bpf_ringbuf_submit` directly, which bypasses the lost-event accounting in `ebpf_ringbuf_write`.

This switches those paths to stack-allocated events plus `ebpf_ringbuf_write`.

Fixes #288

## Test plan
- [ ] Build/load the updated probes on a test host
- [ ] Confirm network/process events still arrive and ring buffer stats look sane under load

### DIFF
--- a/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Network/Probe.bpf.c
@@ -80,20 +80,16 @@ static int inet_csk_accept__exit(struct sock *sk)
     if (ebpf_events_is_trusted_pid())
         goto out;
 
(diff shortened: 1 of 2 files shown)


nicholasberlin:
Thanks for your contribution!!

I'm not sure how to allow CI, so I need to follow up with a PR to enable that.

In the mean time, I copied your PR here: https://github.com/elastic/quark/pull/403
with one minor change to zero the stack variables.

PR #404, opened 2026-08-31
Executable mprotect events via LSM hook, deduplicated per process life

nicholasberlin:
## Summary

Adds executable `mprotect()` attempt events (`QUARK_EV_MPROTECT`, `quark-mon -u`), collected at the `security_file_mprotect` LSM hook and deduplicated to one event per unique protection transition per process life.

This supersedes #400 (same LSM approach, rebuilt on current main rather than on top of #367) and incorporates what we learned comparing #400 against #367 on identical GCP VMs.

### Design

- **LSM hook, per VMA**: the probe sees the previous, requested, and kernel-adjusted effective protection (so `READ_IMPLIES_EXEC` is visible) plus backing-file identity, before the change is committed. Covers `mprotect(2)`, `pkey_mprotect(2)`, and compat callers.
- **Deduplication**: an LRU map keyed by tgid plus a 7-bit transition id (prev prot, effective prot, file-backedness) is claimed with `BPF_NOEXIST`, so concurrent threads of one process race to exactly one winner. Only a confirmed "already reported" suppresses; any map failure emits — the failure mode is a duplicate event, never a lost one. State clears on `execve` and process exit; LRU eviction is a safe backstop for the same reason.
- **File-backed paths**: file-backed attempts are rare and high signal (library tampering, packers, memfd fake JITs), so they carry the mount-namespace relative path as a variable-length field. `quark_mprotect.path` is NULL for anonymous mappings.
- **Stats**: `quark_queue_stats` gains `{submitted, suppressed, reserve_failed}`, shown by `quark-mon -M`, so deployments can observe how much the dedup absorbs.

### Measurements

From a harness that runs identical deterministic workloads on two identical Ubuntu 24.04 VMs (this branch vs. the `sys_enter/exit_mprotect` tracepoint approach of #367):

| workload | executable attempts | events emitted (this PR) | events emitted (#367) |
|---|---|---|---|
| Firefox headless, 5 min deterministic browse | 24,922 | **1** (anon RW→X JIT signature) | 25,389 |
| node JIT workload | 34 | 4 | 34 |
| JVM JIT workload | 2 | 2 | 2 |
| file-backed/memfd/anon mprotect exerciser | 450 | 9 | 450 |

A throwaway instrumentation build also correlated every hook-side attempt with its syscall exit across all workloads: 100% of executable attempts seen at the LSM hook committed (e.g. Firefox 24,919/24,919 succeeded, 0 failed), so hook-side "attempt" events are not inflated by failing calls in practice.

### Tests

`t_mprotect` extended for the new semantics: repeated transitions must not re-emit (sentinel check after flipping protections back and forth), the `pkey_mprotect` case uses R→X so it is distinct from the earlier RW→X, and the file-backed event must carry the path of `/proc/self/exe`. Passes on the ebpf backend (Ubuntu 24.04, 6.14 gcp kernel).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -178,3 +178,10 @@ Changes from 0.8 to 0.9
     previous/requested/effective protection and backing-file identity. These
     events cover mprotect(), pkey_mprotect(), and compat callers, but do not
     assert that the protection change ultimately succeeded.
+    Only the first occurrence of each protection transition (previous
+    protection, effective protection, file-backedness) is reported per process
+    life; the state resets on execve() and process exit. File-backed attempts
+    additionally carry the mount-namespace relative path in
+    quark_mprotect.path, NULL for anonymous mappings.
+  o quark_queue_stats{} got a quark_mprotect_stats member (submitted,
+    suppressed, reserve_failed), also shown by quark-mon -M.
--- a/CHANGES
+++ b/CHANGES
@@ -172,3 +172,9 @@ Changes from 0.8 to 0.9
     CLOCK_BOOTTIME. This effectively drops support for kernels without perf
     clockid support (< 4.1 or RHEL before 7.4).
   o quark_queue_stats{} got a "stalls" member.
+  o Added executable mprotect() attempt events (QUARK_EV_MPROTECT, -u on
+    quark-mon). The eBPF probe reports each VMA that reaches the
+    security_file_mprotect() check with effective execute permission, including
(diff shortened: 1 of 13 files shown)


nicholasberlin:
Split into a stack of five smaller PRs for review: #409 (core LSM event, private), #410 (deduplication, goes public, carries the measurements from this description), #411 (stats), #412 (file-backed path), #413 (Go bindings).

PR #403, opened 2026-08-31
Fix/probes ebpf ringbuf write zero init

nicholasberlin:
Several fixed-size network and process probe paths still called `bpf_ringbuf_reserve`/`bpf_ringbuf_submit` directly, which bypasses the lost-event accounting in `ebpf_ringbuf_write`.

This switches those paths to stack-allocated events plus `ebpf_ringbuf_write`.

Also, zeroes stack variables.

Fixes #288

## Test plan
- [x] Build/load the updated probes on a test host
- [x] Confirm network/process events still arrive and ring buffer stats look sane under load

PR #405, opened 2026-08-31
ebpf: add zeroed event buffer helper

nicholasberlin:
## Summary

- add `get_zeroed_event_buffer()` to combine scratch-buffer lookup and zero-initialization
- use it at the seven fixed-size event call sites introduced by #403
- leave existing variable-length buffer users unchanged

Follow-up to https://github.com/elastic/quark/pull/403#discussion_r3897512808

## Testing

- compiled `bpf_probes.c` for the BPF target with Homebrew Clang 22.1.6
- `git diff --check`


christos68k (approved):
LGTM 

Hopefully it passes verifier

PR #406, opened 2026-08-31
ebpf: compare bpf_probe_read_kernel_str returns in 32 bits

nicholasberlin:
## Summary

Kernels before 5.9 (bdb7b79b4ce8) declared BPF helpers as returning `int`, and the JIT zero-extends the 32-bit result, so 64-bit comparisons against negative errnos silently never match. #404 fixed this for the mprotect dedup `bpf_map_update_elem()` check; this PR fixes the three remaining latent instances, all around `bpf_probe_read_kernel_str()`:

- `GPL/Events/Helpers.h` `read_kernel_str_or_empty_str()`: the empty-string fallback never triggered on old kernels, so a failed read returned a bogus huge value as the string length, which many callers (module_load, symlink/rename, cgroup paths, comm) pass to `ebpf_vl_field__set_size()`.
- `GPL/Events/Process/Probe.bpf.c` `emit_memfd_create_event()`: same shape for the memfd filename read (`long size; ... if (size <= 0)`).
- `GPL/Events/PathResolver.h` `ebpf_resolve_kernfs_node_to_string()`: a failed component read was not caught, leaving garbage in the path buffer.

These only misbehave when `bpf_probe_read_kernel_str()` actually fails on a pre-5.9 kernel (rare, which is why CI never caught them). Worst case is a dropped event or garbage path, not corruption. Fix is the same pattern as #404: do the comparison in 32 bits (`int ret = ...` / `(int)` cast).

Not touched: `nova.bpf.c` has the same shape but requires `bpf_loop` (kernel 5.17+), so it cannot run on affected kernels.

### DIFF
--- a/elastic-ebpf/GPL/Events/Helpers.h
+++ b/elastic-ebpf/GPL/Events/Helpers.h
@@ -146,7 +146,11 @@ static bool IS_ERR_OR_NULL(const void *ptr)
 // Wrapper around bpf_probe_read_kernel_str that reads an empty string upon a read failure
 static long read_kernel_str_or_empty_str(void *dst, int size, const void *unsafe_ptr)
 {
-    long ret = bpf_probe_read_kernel_str(dst, size, unsafe_ptr);
+    // Compare in 32 bits: before Linux 5.9 (bdb7b79b4ce8) helpers were
+    // declared to return int, and the JIT zero-extends the 32-bit result, so
+    // a 64-bit compare against a negative errno never matches on older
+    // kernels.
+    int ret = bpf_probe_read_kernel_str(dst, size, unsafe_ptr);
     if (ret < 0) {
         ((char *)dst)[0] = '\0';
         return 1;
--- a/elastic-ebpf/GPL/Events/PathResolver.h
+++ b/elastic-ebpf/GPL/Events/PathResolver.h
@@ -272,7 +272,11 @@ static size_t ebpf_resolve_kernfs_node_to_string(char *buf, struct kernfs_node *
 
         buf[cur & PATH_MAX_INDEX_MASK] = '/';
         cur                            = (cur + 1) & PATH_MAX_INDEX_MASK;
-        if (bpf_probe_read_kernel_str(
+        // Compare in 32 bits ((int) cast): before Linux 5.9 (bdb7b79b4ce8)
+        // helpers were declared to return int, and the JIT zero-extends the
+        // 32-bit result, so a 64-bit `< 0` compare never matches on older
+        // kernels.
+        if ((int)bpf_probe_read_kernel_str(
                 buf + (cur & PATH_MAX_INDEX_MASK),
                 PATH_MAX - cur > KERNFS_NODE_COMPONENT_MAX_LEN ? KERNFS_NODE_COMPONENT_MAX_LEN : 0,
                 (void *)name) < 0)
--- a/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
@@ -541,7 +541,10 @@ static int emit_memfd_create_event(const char *name)
     // Variable length fields
     ebpf_vl_fields__init(&event->vl_fields);
     struct ebpf_varlen_field *field;
-    long size;
+    // Declared int, not long: before Linux 5.9 (bdb7b79b4ce8) helpers were
+    // declared to return int, and the JIT zero-extends the 32-bit result, so
+    // a 64-bit `size <= 0` check never catches failures on older kernels.
+    int size;
 
     // memfd filename
     field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_FILENAME);


nicholasberlin:
## TL;DR
Closing this: @christos68k tested the premise and it does not hold for `bpf_probe_read_kernel_str()`. Lesson: don't trust, verify.

## Bot's Reproduction and Analysis
Wrote a small raw_tracepoint program that calls two helpers with inputs that must fail, records the raw 64-bit `r0` the BPF program sees, and evaluates both the 64-bit compare quark uses today and the 32-bit compare this PR proposes. Booted it under `krun.sh` on five kernels:

| kernel | `bpf_probe_read_kernel_str(NULL)` r0 | `bpf_map_update_elem(BPF_NOEXIST)` r0 |
|---|---|---|
| 5.8.15 (Fedora 33 GA, pre-5.9) | `0xffffffffffffffde` (-ERANGE) | `0x00000000ffffffef` **zero-extended** |
| 5.14.0-362 (RHEL 9.3) | sign-extended | **zero-extended** |
| 6.2.9 (Fedora 38 GA) | sign-extended | **zero-extended** |
| 6.5.6 (Fedora 39 GA) | sign-extended | `0xffffffffffffffef` (-EEXIST) |
| 6.8.0-65 (Ubuntu) | sign-extended | sign-extended |

So a failing `bpf_probe_read_kernel_str()` is sign-extended correctly on every kernel, including pre-5.9, and the existing `long ret ... if (ret < 0)` in `read_kernel_str_or_empty_str()`, `emit_memfd_create_event()` and `ebpf_resolve_kernfs_node_to_string()` already works. There is nothing for this PR to fix.

Why: every helper's in-kernel entry point is the `u64` wrapper generated by `BPF_CALL_x`, so an `int` error is sign-extended before it reaches r0. Upstream bdb7b79b4ce8 (5.9) only changed the UAPI declaration that libbpf's `bpf_helper_defs.h` is generated from; its own commit message says correctness was already preserved by two's complement. I had the mechanism wrong in the description here and in the #404 comment.

The zero-extension is real, but only for the map helpers (`bpf_map_update_elem`, `delete`, `push`, `pop`, `peek`): with the JIT enabled, the verifier's `fixup_bpf_calls` rewrites those calls into direct calls to the map ops, bypassing the wrapper, and the ops returned `int` until d7ba4cc900bf ("bpf: return long from bpf_map_ops funcs", Linux 6.4). That is exactly the `-EEXIST` compare in the #404 dedup path, so the `(int)` cast there stays, and I have corrected its comment to cite the right commit and the right kernel range (< 6.4, not < 5.9). The only other map-helper return check in quark's BPF code is the `!= 0` test on `bpf_map_delete_elem` in the network probe, which is immune to zero-extension.

<details>
<summary>Reproducer (build from a quark checkout, run under krun.sh on the Fedora 33 GA 5.8.15 kernel)</summary>

`ret32.bpf.c`:
```c
#include "vmlinux.h"
#include <bpf/bpf_helpers.h>

char LICENSE[] SEC("license") = "GPL";

struct {
      __uint(type, BPF_MAP_TYPE_ARRAY);
      __uint(max_entries, 8);
      __type(key, __u32);
      __type(value, __u64);
} results SEC(".maps");

struct {
      __uint(type, BPF_MAP_TYPE_LRU_HASH);
      __uint(max_entries, 4);
      __type(key, __u64);
      __type(value, __u8);
} seen SEC(".maps");

static __always_inline void put(__u32 i, __u64 v)
{
      __u64 *p = bpf_map_lookup_elem(&results, &i);
      if (p)
              *p = v;
}

SEC("raw_tracepoint/sys_enter")
int sys_enter(void *ctx)
{
      char    buf[16];
      long    ret;
      __u64   key = 1;
      __u8    one = 1;

      ret = bpf_probe_read_kernel_str(buf, sizeof(buf), (void *)0);
      put(0, ret);                    /* raw r0 */
      put(1, ret < 0);                /* 64-bit compare (current quark code) */
      put(2, (int)ret < 0);           /* 32-bit compare (this PR) */

      bpf_map_update_elem(&seen, &key, &one, BPF_ANY);
      ret = bpf_map_update_elem(&seen, &key, &one, BPF_NOEXIST);
      put(3, ret);                    /* raw r0 */
      put(4, ret == -17);             /* 64-bit compare */
      put(5, (int)ret == -17);        /* 32-bit compare (#404 dedup) */

      put(7, 1);
      return 0;
}
```

`ret32.c`:
```c
#include <errno.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>
#include <sys/utsname.h>
#include <bpf/libbpf.h>

int
main(int argc, char *argv[])
{
      const char              *path = argc > 1 ? argv[1] : "ret32.bpf.o";
      struct bpf_object       *obj;
      struct bpf_program      *prog;
      struct bpf_link         *link;
      struct bpf_map          *map;
      struct utsname           uts;
      __u64                    v[8] = {0};
      __u32                    i;
      int                      err, bad = 0;

      uname(&uts);
      printf("kernel %s\n", uts.release);

      obj = bpf_object__open_file(path, NULL);
      if (!obj) { fprintf(stderr, "open %s: %s\n", path, strerror(errno)); return 2; }
      if ((err = bpf_object__load(obj))) { fprintf(stderr, "load: %s\n", strerror(-err)); return 2; }
      prog = bpf_object__find_program_by_name(obj, "sys_enter");
      link = bpf_program__attach(prog);
      if (!link) { fprintf(stderr, "attach: %s\n", strerror(errno)); return 2; }

      getpid();

      map = bpf_object__find_map_by_name(obj, "results");
      for (i = 0; i < 8; i++)
              bpf_map__lookup_elem(map, &i, sizeof(i), &v[i], sizeof(v[i]), 0);
      bpf_link__destroy(link);

      if (!v[7]) { fprintf(stderr, "program did not run\n"); return 2; }

      printf("bpf_probe_read_kernel_str(NULL):  r0=0x%016llx (%lld)  ret<0 [64-bit]=%llu  (int)ret<0 [32-bit]=%llu\n",
          v[0], (long long)v[0], v[1], v[2]);
      printf("bpf_map_update_elem(BPF_NOEXIST): r0=0x%016llx (%lld)  ret==-EEXIST [64-bit]=%llu  (int)ret==-EEXIST [32-bit]=%llu\n",
          v[3], (long long)v[3], v[4], v[5]);

      if (!v[1] || !v[4])
              bad = 1;
      printf("%s: 64-bit compare of helper return %s on this kernel\n",
          bad ? "FAIL" : "OK", bad ? "MISSES negative errno" : "works");
      return bad;
}
```

```sh
clang-20 -g -O2 -target bpf -Ielastic-ebpf/contrib/vmlinux/x86_64 -Iinclude -c ret32.bpf.c -o ret32.bpf.o
cc -O2 -static -Iinclude ret32.c -o ret32 libbpf/src/libbpf.a elftoolchain/libelf/libelf_pic.a zlib/libz.a
mkdir -p rootfs && cp init ret32 ret32.bpf.o rootfs/
(cd rootfs && find . -print0 | cpio -0 -o --format=newc | gzip -9 > ../initramfs-ret32.gz)
curl -sO https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/33/Everything/x86_64/os/Packages/k/kernel-core-5.8.15-301.fc33.x86_64.rpm
rpm2cpio kernel-core-5.8.15-301.fc33.x86_64.rpm | cpio -idm './lib/modules/*/vmlinuz'
./krun.sh initramfs-ret32.gz lib/modules/5.8.15-301.fc33.x86_64/vmlinuz /ret32 /ret32.bpf.o
```

Output on 5.8.15:
```
bpf_probe_read_kernel_str(NULL):  r0=0xffffffffffffffde (-34)  ret<0 [64-bit]=1  (int)ret<0 [32-bit]=1
bpf_map_update_elem(BPF_NOEXIST): r0=0x00000000ffffffef (4294967279)  ret==-EEXIST [64-bit]=0  (int)ret==-EEXIST [32-bit]=1
FAIL: 64-bit compare of helper return MISSES negative errno on this kernel
```
</details>

PR #407, opened 2026-09-01
Added a CPU count call and determine ring buffer size.

matthewscherer:
Size the Linux eBPF process-event ring buffer at load from CPU count instead of the fixed 4 MiB map.

Before probes load, set max_entries to 512 KiB × ncpu, rounded up to a power of two, floored at 4 MiB and capped at 64 MiB. That keeps the per-CPU headroom the old 4 MiB default had on an 8-CPU box (a 128-CPU host gets 64 MiB). Same formula in quark (bpf_queue.c) and the Kernel/Ebpf fallback.

### DIFF
--- a/bpf_probes.c
+++ b/bpf_probes.c
@@ -6,7 +6,8 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
-	__uint(max_entries, 1 << 22); // 4 MiB
+	/* 4 MiB compile-time default; userspace resizes at load from ncpu. */
+	__uint(max_entries, 1 << 22);
 } ringbuf SEC(".maps");
 
 #include "Process/Probe.bpf.c"
--- a/bpf_probes.c
+++ b/bpf_probes.c
@@ -6,8 +6,8 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
-	/* 4 MiB compile-time default; userspace resizes at load from ncpu. */
-	__uint(max_entries, 1 << 22);
+	/* The size is set in userspace derived from ncpu. */
+	__uint(max_entries, 0);
 } ringbuf SEC(".maps");
 
 #include "Process/Probe.bpf.c"
--- a/bpf_queue.c
+++ b/bpf_queue.c
@@ -4,7 +4,7 @@
 #include <sys/epoll.h>
 #include <sys/param.h>
 #include <sys/mount.h>
-#include <sys/sysinfo.h>
+#include <sys/resource.h>
 
 #include <assert.h>
 #include <ctype.h>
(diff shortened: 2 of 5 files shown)


nicholasberlin (commented):
Code looks good to me. Needs an entry in CHANGES.

matthewscherer:
> Code looks good to me. Needs an entry in CHANGES.

Updated.

bpf_queue.c
```diff
@@ -1286,12 +1330,19 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
 	if (qq->flags & QQ_GETPID)
 		bpf_program__set_autoload(p->progs.tracepoint_syscalls_sys_exit_getpid, 1);
 
-	if (bpf_map__set_max_entries(p->maps.event_buffer_map,
-	    get_nprocs_conf()) != 0) {
+	ncpu = get_nprocs_conf();
```
christos68k: We probably need to use `libbpf_num_possible_cpus()` here, same as in [bpf_queue_update_stats](https://github.com/elastic/quark/blob/d8405f923170ca5414803aa53a254737d9c50521/bpf_queue.c#L1467) to take into account the divergence between possible CPUs and online CPUs and other gotchas. This is also what the kernel is using for per-CPU maps.

**EDIT:** Let me double check what `get_nprocs_conf` is returning and how it's derived.

bpf_queue.c
```diff
@@ -1286,12 +1330,19 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
...
+	if (bpf_map__set_max_entries(p->maps.event_buffer_map, ncpu) != 0) {
 		qwarn("bpf_map__set_max_entries event_buffer_map");
 		goto fail;
 	}
 
+	ringbuf_size = bpf_ringbuf_size(ncpu);
+	if (bpf_map__set_max_entries(p->maps.ringbuf, ringbuf_size) != 0) {
```
christos68k, line 1374: This will probably fail on older kernels due to `RLIMIT_MEMLOCK`. Typically, we'd bump this limit before resizing maps (e.g. see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/06ea040c39d3d17bc1534a5dcc044368caf48782/tracer/tracer.go#L657-L658)). I'm not sure if this is happening behind-the-scenes in quark due to libbpf. 

But ideally, we also need a test that tries for a maximally sized ringbuf as that will manifest the problem on all the kernels we have at buildkite.

matthewscherer: A test was written here. https://github.com/elastic/quark/pull/407/changes/ab28324e23be13655f0d78ec8b8a8f4bb0e330d5

bpf_queue.c
christos68k: Ok so `get_nprocs_conf` is a libc extension that essentially returns possible cpus but it's probably better to use `libbpf_num_possible_cpus` which is more explicit.

bpf_probes.c
```diff
@@ -6,7 +6,8 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
-	__uint(max_entries, 1 << 22); // 4 MiB
+	/* 4 MiB compile-time default; userspace resizes at load from ncpu. */
+	__uint(max_entries, 1 << 22);
```
christos68k: We should set this to zero to avoid the overhead of the initial allocation which is thrown away (this is a common pattern with BPF maps that are sized at load-time).

```suggestion
	/* The size is set in userspace derived from ncpu. */
	__uint(max_entries, 0);
```

christos68k (dismissed):
LGTM (with maybe some follow-up work on sizing if we think that 64MiB is excessive for systems with large difference between online and possibe CPUs)

nicholasberlin (approved):

PR #409, opened 2026-09-02
Collect executable mprotect attempts at LSM hook

nicholasberlin:
Part 1 of 4 of the executable `mprotect()` event work, split out of #404 (which superseded #400) so each piece can be reviewed on its own. The stack, bottom-up, each PR based on the one before it:

1. this PR: core LSM event, private
4. #413 Go bindings

## Summary

Adds `QUARK_EV_MPROTECT` and `QQ_MPROTECT`. The eBPF probe reports each VMA that reaches the `security_file_mprotect` LSM hook with effective execute permission, carrying the previous, requested and kernel-adjusted effective protection (so `READ_IMPLIES_EXEC` is visible) plus backing-file identity (inode, dev). It covers `mprotect(2)`, `pkey_mprotect(2)` and compat callers. The hook runs before the change is committed, so an event is an attempt, not a completed change.

This PR is deliberately private in the `QQ_GETPID` sense: the flags exist in `quark.h`, marked experimental, so `t_mprotect` can drive the probe, but there is no `quark-mon` flag, manual page or CHANGES entry. Those land with the deduplication in #410, so `main` never carries a public, per-call noisy mprotect event. Measurements motivating the deduplication are in #410.

Rebuilt from #404 on top of the zeroed event buffer helper (#405); the fixed-size event uses `get_zeroed_event_buffer()` like the other fixed-size events. `QQ_MPROTECT` takes bit 15, freed by the `QQ_MONOTONIC` removal.

## Tests

`t_mprotect` (ebpf backend): non-executable protections and a rejected unaligned call must not emit; the four anonymous RW→{X, RX, WX, RWX} transitions each emit exactly once with the expected previous, requested and effective protection; the file-backed event carries the inode and dev of `/proc/self/exe`; `pkey_mprotect` is observed through the same hook. Passes locally on Ubuntu 6.8 (fentry).

### DIFF
--- a/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
@@ -514,10 +514,6 @@ static int security_file_mprotect__enter(struct vm_area_struct *vma,
     event->prev_prot       = prev_prot;
     event->req_prot        = req_prot;
     event->effective_prot  = effective_prot;
-    event->file_backed     = 0;
-    event->inode           = 0;
-    event->dev_major       = 0;
-    event->dev_minor       = 0;
 
     if (file) {
         event->file_backed = 1;
--- a/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
@@ -25,6 +25,17 @@ DECL_FIELD_OFFSET(iov_iter, __iov);
 #define S_ISUID 0004000
 #define S_ISGID 0002000
 
+#define MPROTECT_PROT_READ  0x1
+#define MPROTECT_PROT_WRITE 0x2
+#define MPROTECT_PROT_EXEC  0x4
+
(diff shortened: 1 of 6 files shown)


elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -450,6 +461,116 @@ int BPF_KPROBE(kprobe__arch_ptrace,
...
+
+    event->vma_start       = BPF_CORE_READ(vma, vm_start);
+    event->vma_end         = BPF_CORE_READ(vma, vm_end);
+    event->prev_prot       = prev_prot;
+    event->req_prot        = req_prot;
+    event->effective_prot  = effective_prot;
+    event->file_backed     = 0;
+    event->inode           = 0;
+    event->dev_major       = 0;
+    event->dev_minor       = 0;
```
christos68k: We can skip these as the event buffer is zeroed.

```suggestion
```

christos68k (dismissed):
LGTM

christos68k (approved):

PR #410, opened 2026-09-02
Deduplicate mprotect events per process life and make them public

nicholasberlin:
Part 2 of 4 of the executable `mprotect()` event work, split out of #404 (which superseded #400) so each piece can be reviewed on its own. The stack, bottom-up, each PR based on the one before it:

2. this PR: deduplicate per process life, make public (`quark-mon -u`, manual pages, CHANGES)
4. #413 Go bindings

## Summary

One event per unique protection transition per process life. An LRU map keyed by tgid holds a one-word bitmap of the 64 possible transitions: previous protection (3 bits), read/write bits of the effective new protection (2 bits, execute is always set by the time the dedup runs) and file-backedness (1 bit). The entry is created empty with `BPF_NOEXIST` so a racing thread can't clobber bits already set, and every transition then tests and sets its bit with a plain read-modify-write (an atomic or needs Linux 5.12 and `-mcpu=v3`). Best effort by design: two threads of one process racing on the same transition can both report, and any map failure reports, so the failure mode is a duplicate event, never a lost one. The entry is dropped on `execve` (new program image) and on process exit (tgid reuse), one delete each; LRU eviction is a safe backstop for the same reason.

The feature costs nothing when `QQ_MPROTECT` is off: the exec and exit bookkeeping sits behind a rodata flag the verifier prunes, and the map, declared with zero entries, is sized from userspace only when enabled (16384 processes, about 1 MB) and otherwise not created at all, following the pattern #407 set for the ring buffer.

With duplicates gone the event is fit for consumption, so this PR also makes it public: the experimental markers on `QQ_MPROTECT`/`QUARK_EV_MPROTECT` go away, `quark-mon` gets `-u`, `quark_queue_open(3)` and `quark_queue_get_event(3)` document the flag and event, and CHANGES gets its entry.

The second commit is @christos68k's simplification of the dedup (one word instead of two, no `-EEXIST` compare), cherry-picked from his draft branch.

## Measurements

Carried over from #404. From a harness that runs identical deterministic workloads on two identical Ubuntu 24.04 GCP VMs (this approach vs. the `sys_enter/exit_mprotect` tracepoint approach of #367):

| workload | executable attempts | events emitted (LSM hook + dedup) | events emitted (#367) |
|---|---|---|---|
| Firefox headless, 5 min deterministic browse | 24,922 | **1** (anon RW→X JIT signature) | 25,389 |
| node JIT workload | 34 | 4 | 34 |
| JVM JIT workload | 2 | 2 | 2 |
| file-backed/memfd/anon mprotect exerciser | 450 | 9 | 450 |

A throwaway instrumentation build also correlated every hook-side attempt with its syscall exit across all workloads: 100% of executable attempts seen at the LSM hook committed (e.g. Firefox 24,919/24,919 succeeded, 0 failed), so hook-side "attempt" events are not inflated by failing calls in practice.

## Tests

`t_mprotect`: a repeated RW→X transition on the same mapping must not re-emit (the file-backed event drained afterwards doubles as the sentinel), and the `pkey_mprotect` case now uses R→X so it is not deduplicated against the earlier RW→X. Passes locally on 6.8 (Ubuntu), 5.14 (RHEL 9.3) and 5.8.15 (Fedora 33), and the full suite passes on 6.8, which also covers the disabled path since every other ebpf test loads the exec and exit probes with the map not created.

### DIFF
--- a/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
@@ -36,6 +36,92 @@ DECL_FIELD_OFFSET(iov_iter, __iov);
 #define MPROTECT_MINORBITS 20
 #define MPROTECT_MINORMASK ((1U << MPROTECT_MINORBITS) - 1)
 
+#ifndef EEXIST
+#define EEXIST 17 /* uapi asm-generic/errno-base.h; not carried by vmlinux.h */
+#endif
+
+// Set from userspace when QQ_MPROTECT is enabled. Everything that touches
+// mprotect_seen from the always loaded exec and exit probes is behind this
+// flag, so with the feature off the verifier prunes it and the map is not
+// even created.
+const volatile bool mprotect_enabled = false;
+
+/*
+ * Executable mprotect transitions already reported for a process life, one
+ * entry per tgid. A transition id packs the previous protection (3 bits), the
+ * effective new protection (3 bits) and whether the mapping is file backed
+ * (1 bit), so all 128 fit a two word bitmap. The entry is dropped when the
+ * process execs or dies (each program image starts fresh); LRU eviction is a
+ * safe backstop since an evicted entry only causes a duplicate event, never a
+ * lost one. Sized from userspace, see bpf_queue_open1().
+ */
+struct mprotect_seen_bits {
+    u64 bits[2];
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u32);
+    __type(value, struct mprotect_seen_bits);
+    __uint(max_entries, 0);
+} mprotect_seen SEC(".maps");
+
+static u32 mprotect_transition_id(u64 prev_prot, u64 effective_prot, u64 file_backed)
+{
+    return ((prev_prot & 0x7) << 4) | ((effective_prot & 0x7) << 1) | (file_backed & 0x1);
+}
+
+// Returns true if the transition was already reported for this process life.
+// The first transition of a life creates the entry with BPF_NOEXIST so
+// concurrent threads racing for it get exactly one winner. Only a confirmed
+// "already reported" suppresses -- any other failure reports, so the failure
+// mode is a duplicate event, never a lost one.
+static bool mprotect_seen__test_and_set(u32 tgid, u32 id)
+{
+    u32 word = (id >> 6) & 1;
+    u64 bit  = 1ULL << (id & 63);
+
+    struct mprotect_seen_bits *seen = bpf_map_lookup_elem(&mprotect_seen, &tgid);
+    if (seen == NULL) {
+        // Constant stack offsets on purpose: indexing the stack copy by word
+        // is a variable offset stack access, which kernels before 5.13 reject.
+        struct mprotect_seen_bits fresh = {
+            .bits = { word ? 0 : bit, word ? bit : 0 },
+        };
+
+        // Compare in 32 bits: with the JIT enabled, the verifier rewrites map
+        // helper calls into direct calls to the map ops (fixup_bpf_calls), and
+        // those returned int until Linux 6.4 (d7ba4cc900bf), leaving a
+        // zero-extended 32-bit value in r0 on older kernels. A 64-bit compare
+        // against -EEXIST never matches there. Ordinary helpers return through
+        // a u64 wrapper and are sign-extended correctly; only map helpers need
+        // this.
+        if ((int)bpf_map_update_elem(&mprotect_seen, &tgid, &fresh, BPF_NOEXIST) != -EEXIST)
+            return false;
+        // Lost the race for the entry, test the bit in the winner's.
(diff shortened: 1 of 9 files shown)


elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -36,6 +36,44 @@ DECL_FIELD_OFFSET(iov_iter, __iov);
...
+ * mapping is file backed (1 bit), so at most 128 entries per process. Entries
+ * are cleared when the process execs or dies (each program image starts
+ * fresh); LRU eviction is a safe backstop since an evicted entry only causes
+ * a duplicate event, never a lost one.
+ */
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u64);
+    __type(value, u8);
+    __uint(max_entries, 65536);
```
christos68k: Since this feature is optional, we can set the map size to zero and only allocate at map-load time if we enable the probes. That way we don't pay the memory cost which here would be a few MB.


```suggestion
    __uint(max_entries, 0);
```

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -36,6 +36,44 @@ DECL_FIELD_OFFSET(iov_iter, __iov);
...
+{
+    u64 id = ((prev_prot & 0x7) << 4) | ((effective_prot & 0x7) << 1) | (file_backed & 0x1);
+    return ((u64)tgid << 8) | id;
+}
+
+static void mprotect_seen__clear(u32 tgid)
+{
+    u64 base = (u64)tgid << 8;
+
+    for (u32 i = 0; i < MPROTECT_TRANSITION_MAX; i++) {
```
christos68k: We'll pay this map_delete*128 cost twice per process execution, even if the probes are not loaded which adds latency to each process exec. We can do better here, some options:

1. Add a bool flag (e.g. `mprotect_enabled`) in BPF `rodata` that we set from userspace. Then any logic that exists in a branch on that flag that's not taken can be removed by the verifier so there's no runtime cost at all if mprotect isn't enabled. That's what we do in the profiler (see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/069c1e2ad85e8323aa75f0b10447f90bf8b5d5d0/support/ebpf/native_stack_trace.ebpf.c#L6-L28) and [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/069c1e2ad85e8323aa75f0b10447f90bf8b5d5d0/tracer/systemconfig.go#L566)). Quark is using libbpf so maybe the mechanism to do this is different but should still be possible.
2. Store the 128 possible values as a bitmap composed of two contatenated 64bit integers. Then clearing the process state is one delete op instead of 128. There's [__sync_fetch_and_or](https://docs.ebpf.io/linux/concepts/concurrency/) on 5.12+, here we'd need a racy fallback that reads, tests and writes the full value. Maybe concurrent writes are not problem as the event wouldn't be dropped (just opens possibility of future duplicate).
3. Add a globally unique process lifetime id to the key (e.g. https://github.com/elastic/quark/blob/12dc26dbd37cf90d8eced342c5ff1949b004eeee/elastic-ebpf/GPL/Events/Helpers.h#L273) and let the LRU expire the entry.
4. ~Do batch deletes in userspace (adds races, not possible on older kernels, mentioning this for posterity and it's also something we do on the profiler)~

nicholasberlin: went with options 1 and 2 combined. thanks!

commit: https://github.com/elastic/quark/pull/410/commits/b3f9417f3973219cddf7352dcfb6211156a5e890
compare link: compare link : https://github.com/elastic/quark/compare/12dc26dbd37cf90d8eced342c5ff1949b004eeee..b3f9417f3973219cddf7352dcfb6211156a5e890

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
nicholasberlin: Done, and a bit further.
commit: https://github.com/elastic/quark/pull/410/commits/b3f9417f3973219cddf7352dcfb6211156a5e890.
compare link : https://github.com/elastic/quark/compare/12dc26dbd37cf90d8eced342c5ff1949b004eeee..b3f9417f3973219cddf7352dcfb6211156a5e890

The map is now declared with `max_entries 0` and set in `bpf_queue_open1()` when `QQ_MPROTECT` is set.

Further, when the flag is off we call `bpf_map__set_autocreate(map, false)`, which stops the map from being created at all. The exec and fork probes only access the map if the rodata flag (from your comment below) is set, so no worries there.

The map's key->value changed to tgid->128 bits value, so the max entries was reduced to 16K. 16K concurrent process adding X via mprotect feels like enough to me. Agreed?

christos68k (approved):
Thanks, LGTM

PR #411, opened 2026-09-02
Add mprotect deduplication counters to quark_queue_stats

nicholasberlin:
Part 3 of 5 of the executable `mprotect()` event work, split out of #404 (which superseded #400) so each piece can be reviewed on its own. The stack, bottom-up, each PR based on the one before it:

3. this PR: deduplication counters in `quark_queue_stats`
4. #412 backing-file path on file-backed events
5. #413 Go bindings

## Summary

`quark_queue_stats{}` gains a `quark_mprotect_stats` member with `submitted`, `suppressed` and `reserve_failed`, fed by a per-CPU array the probe bumps on every ring buffer write, dedup suppression and buffer failure. It is read through `quark_queue_get_stats(3)` and shown by `quark-mon -M` when `-u` is set, so deployments can observe how much the deduplication absorbs.

## Tests

`t_mprotect` now asserts that the counters reflect the transitions it drove: at least the four anonymous submissions plus the file-backed one, and at least one suppression from the repeated RW→X. This assertion is new relative to #404, which did not exercise the counters. Passes locally on Ubuntu 6.8 (fentry).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -181,3 +181,5 @@ Changes from 0.8 to 0.9
     Only the first occurrence of each protection transition (previous
     protection, effective protection, file-backedness) is reported per process
     life; the state resets on execve() and process exit.
+  o quark_queue_stats{} got a quark_mprotect_stats member (submitted,
+    suppressed, reserve_failed), also shown by quark-mon -M.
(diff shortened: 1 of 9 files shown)


nicholasberlin:
Dropped: the deduplication counters are not needed. #412 (path) is now based directly on #410, and #413 no longer adds MprotectStats.

PR #412, opened 2026-09-02
Carry the backing-file path on file-backed mprotect events

nicholasberlin:
Part 3 of 4 of the executable `mprotect()` event work, split out of #404 (which superseded #400) so each piece can be reviewed on its own. The stack, bottom-up, each PR based on the one before it:

3. this PR: backing-file path on file-backed events
4. #413 Go bindings

## Summary

File-backed executable attempts are rare and high signal (library tampering, packers, memfd fake JITs), so they now carry the mount-namespace relative path resolved from the VMA's file, as a variable-length field. The event keeps the zeroed fixed-size buffer (so the file identity fields stay zero for anonymous mappings, #409 dropped the explicit zeroing for that reason) and appends the variable-length tail after it. Anonymous mappings carry no path: `quark_mprotect.path` is `NULL`. The path is owned by the event storage and freed on the next `quark_queue_get_event(3)`.

**Second commit: per-file deduplication.** With the transition alone as the key, a process that makes two different files executable the same way (say RX→RWX on two libraries) reported only the first, which is the very case the file-backed event exists for (raised by @christos68k; per-file semantics chosen with @Aegrah for the lower false-negative rate). File-backed transitions now go through a second LRU map keyed by tgid, device and inode, so each file is reported once per transition per program image (per file per process life, the key includes the tgid) while anonymous mappings keep the per-process bitmap from #410. The value carries the process start time and `self_exec_id`, which the kernel bumps on every exec and never reuses within a process life: an entry left by a dead process (recycled tgid) or a previous program image no longer matches and is claimed anew, so a stale entry never suppresses and the map needs no exec or exit clearing. (The `mm` pointer would not do: it is freed at exec and can come straight back from the slab two images later.) LRU eviction handles leftovers and at worst costs a duplicate. Sized from userspace (16384 files, about 1.5 MB) and not created when `QQ_MPROTECT` is off. The volume argument stays intact: anonymous JIT churn is still collapsed per process, and the file-backed count is bounded by the number of distinct files a process makes executable.

**Third commit** is @christos68k's fix for RHEL 8: that kernel keeps kABI by renaming the original u32 `self_exec_id` and moving the live u64 counter into `task_struct_rh`, so the read goes through `bpf_core_field_exists` guards with `___el8` partial types in vmlinux_extra.h (cherry-picked from his draft branch).

## Tests

`t_mprotect` checks that the file-backed event carries the path of `/proc/self/exe`, that a repeated R→RX on that file is not reported, and that the same R→RX on a second file (`true`) is reported with that file's inode, device and path. Passes locally on 6.8, 5.14 (RHEL 9.3), 5.8.15 and RHEL 8's 4.18.0-553; the full `quark-test` suite was also run on 6.8 and on the RHEL 8 kernel at this stage.

### DIFF
--- a/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
@@ -131,16 +131,24 @@ struct {
     __uint(max_entries, 0);
 } mprotect_file_seen SEC(".maps");
 
-// The exec generation of the process. self_exec_id is u64 since Linux 5.7,
-// which every kernel this ring buffer backend loads on has, but read it by its
-// relocated size anyway; the targets are little endian so a narrower field
-// lands in the low bytes.
+// The exec generation of the process. self_exec_id is u64 since Linux 5.7 and
+// u32 before, so read it by its relocated size; the targets are little endian
+// so a narrower field lands in the low bytes. RHEL 8 (4.18) backported the
+// widening under kABI and the live field only exists in the task_struct_rh
+// extension there, see vmlinux_extra.h; a kernel with neither yields zero and
+// falls back to start_time alone.
 static u64 mprotect_exec_id(const struct task_struct *task)
 {
     const struct task_struct *leader = BPF_CORE_READ(task, group_leader);
     u64                       exec_id = 0;
 
-    bpf_core_read(&exec_id, bpf_core_field_size(leader->self_exec_id), &leader->self_exec_id);
+    if (bpf_core_field_exists(leader->self_exec_id)) {
+        bpf_core_read(&exec_id, bpf_core_field_size(leader->self_exec_id),
+                      &leader->self_exec_id);
+    } else if (bpf_core_field_exists(struct task_struct___el8, task_struct_rh)) {
+        exec_id = BPF_CORE_READ((const struct task_struct___el8 *)leader, task_struct_rh,
+                                self_exec_id);
+    }
 
     return exec_id;
 }
--- a/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
+++ b/elastic-ebpf/GPL/Events/Process/Probe.bpf.c
@@ -592,6 +592,8 @@ static int security_file_mprotect__enter(struct vm_area_struct *vma,
     event->req_prot        = req_prot;
     event->effective_prot  = effective_prot;
 
+    ebpf_vl_fields__init(&event->vl_fields);
+
     if (file) {
         event->file_backed = 1;
 
@@ -606,9 +608,18 @@ static int security_file_mprotect__enter(struct vm_area_struct *vma,
                 event->dev_minor = (u32)dev & MPROTECT_MINORMASK;
             }
         }
+
+        // File-backed executable transitions are rare and high signal
+        // (library tampering, packers, memfd JITs); annotate them with the
+        // mount-namespace-relative path. Anonymous memory has no path.
+        struct ebpf_varlen_field *field =
+            ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_PATH);
+        struct path p = BPF_CORE_READ(file, f_path);
+        long size     = ebpf_resolve_path_to_string(field->data, &p, task);
+        ebpf_vl_field__set_size(&event->vl_fields, field, size);
     }
 
-    ebpf_ringbuf_write(&ringbuf, event, sizeof(*event), 0);
+    ebpf_ringbuf_write(&ringbuf, event, EVENT_SIZE(event), 0);
 
 out:
     return 0;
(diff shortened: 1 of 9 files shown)


elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -568,7 +568,7 @@ static int security_file_mprotect__enter(struct vm_area_struct *vma,
                                  BPF_NOEXIST) == -EEXIST)
         goto out;
 
-    struct ebpf_process_mprotect_event *event = get_zeroed_event_buffer(sizeof(*event));
```
christos68k, line 581: We can keep this and remove the explicit field sets to 0 later, variable length events still work. This guarantees all fields in the fixed header are zeroed, without us having to manually set them.

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -600,9 +602,18 @@ static int security_file_mprotect__enter(struct vm_area_struct *vma,
                 event->dev_minor = (u32)dev & MPROTECT_MINORMASK;
             }
         }
+
+        // File-backed executable transitions are rare and high signal
+        // (library tampering, packers, memfd JITs); annotate them with the
+        // mount-namespace-relative path. Anonymous memory has no path.
+        struct ebpf_varlen_field *field =
```
christos68k, line 709: Should we add the inode to the key? Otherwise we're only ever reporting one file per process (if the transition id is the same across multiple files per process).

~That invalidates the [bitmap state alternative](https://github.com/elastic/quark/pull/410#discussion_r3918303439) but we could do the LRU expiration instead.~

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -586,6 +586,8 @@ static int security_file_mprotect__enter(struct vm_area_struct *vma,
     event->dev_major       = 0;
     event->dev_minor       = 0;
```
christos68k:
```suggestion
```

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
nicholasberlin: @Aegrah what do you think here? is one file-backed mapping +x for a single process good? or would like one per file?

Aegrah: To reduce FNs, it would be good to have this on a per-file basis. However, the main reason we moved to deduplication here was to reduce telemetry volume. I think for most detection cases, seeing it just for a unique process would be enough. 

If volume is not a large concern here, I'd prefer to see it per file. If volume is a concern, I am fine with it on a per-process basis.

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
nicholasberlin: This comment is from a before time. This PR has since been updated with a previous PR's commit: https://github.com/elastic/quark/pull/409/changes/bc7bd1c9559e1f47a1a62c7b77b11d141adcb553

TL;DR - done

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
nicholasberlin: done

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
nicholasberlin: ok, went with per file: https://github.com/elastic/quark/pull/412/commits/9e24e5c88d36e4733497d0a439787eae187d5414

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
```diff
@@ -98,6 +98,69 @@ static void mprotect_seen__clear(u32 tgid)
...
+ * leftovers and at worst costs a duplicate. Only referenced from the mprotect
+ * probes, sized from userspace, see bpf_queue_open1().
+ */
+struct mprotect_file_key {
+    u32 tgid;
+    u32 dev;
+    u64 inode;
+};
+
+struct mprotect_file_seen {
```
christos68k, line 121: We could merge this and `mprotect_seen`, then the anonynous mapping case would be covered by the key: `(tgid, 0, 0)`. But maybe it's better to keep the maps separate in order to be able to size them differently and not mix the two LRU expiration behaviors together.

christos68k (dismissed):

elastic-ebpf/GPL/Events/Process/Probe.bpf.c
nicholasberlin: > not mix the two LRU expiration behaviors together.

yeah, let's keep them separate.

nicholasberlin:
@christos68k thanks for the review, but my bot found some issues with the latest commit. Need another reviewing after force pushing a new commit: https://github.com/elastic/quark/compare/9e24e5c88d36e4733497d0a439787eae187d5414..d3bdfb735f2bcffd6f4e5beedaf98cd3ce1427e6

christos68k (approved):
LGTM

PR #413, opened 2026-09-02
go: expose mprotect events

nicholasberlin:
Part 4 of 4 of the executable `mprotect()` event work, split out of #404 (which superseded #400) so each piece can be reviewed on its own. The stack, bottom-up, each PR based on the one before it:

4. this PR: Go bindings

## Summary

Go bindings for the mprotect events: `Mprotect` on `Event` mirroring `struct quark_mprotect` (`Path` is the empty string for anonymous mappings), and the `QQ_MPROTECT` and `QUARK_EV_MPROTECT` constants.

Compile-checked with `make quark-go-test` and `go vet`; the Go test binary needs a root box or the VM initramfs with glibc to run.

### DIFF
--- a/go/quark/quark.go
+++ b/go/quark/quark.go
@@ -183,6 +183,21 @@ type Ptrace struct {
 	Data     uint64
 }
 
+// Mprotect describes an executable protection attempt observed for one VMA.
+// The LSM check happens before the kernel commits the protection change.
+type Mprotect struct {
+	VmaStart      uint64
+	VmaEnd        uint64
+	PrevProt      uint64
+	ReqProt       uint64
+	EffectiveProt uint64
+	Inode         uint64
+	DevMajor      uint32
+	DevMinor      uint32
+	FileBacked    bool
+	Path          string // mount-ns relative; empty for anonymous mappings
+}
+
 type ModuleLoad struct {
 	Name       string
 	Version    string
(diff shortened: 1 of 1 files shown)


go/quark/quark.go
```diff
@@ -778,6 +800,23 @@ func ptraceFromC(cPtrace *C.struct_quark_ptrace) Ptrace {
 	return ptrace
 }
 
+func mprotectFromC(cMprotect *C.struct_quark_mprotect) Mprotect {
```
christos68k, line 803: We could get rid of these manual struct C-Go conversions and duplicate definitions by using `cgo -godefs` to auto-generate Go-compatible structs (e.g. see [here](https://utcc.utoronto.ca/~cks/space/blog/programming/GoCGoCompatibleStructs) and [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/types.go)). This also keeps the C struct memory layout intact.

Not for this PR, maybe something to consider for the future if there's consensus.

christos68k (dismissed):
LGTM

stanek-michal (dismissed):

go/quark/quark.go
nicholasberlin: agreed, worth a look for all the *FromC helpers once this lands

christos68k (approved):

PR #416, opened 2026-09-10
Add process_vm_access events for process_vm_readv/writev

Aegrah:
## Summary

Adds `process_vm_access` events for `process_vm_readv` and `process_vm_writev`, which
read and write another process's memory directly. The event reports the target, the
direction, the iovec shape, the remote address and the raw syscall return, so a denied
or failed attempt stays distinguishable from a completed transfer.

Supersedes #376, which resolved the target from the `pid` argument at syscall entry.
That value is namespace-relative and may name a thread, so it was compared against a
global tgid. This resolves the target from the task `mm_access()` has already looked up.

Issue: Prompt-driven feature

- Hook/backend: `tracepoint/syscalls/sys_{enter,exit}_process_vm_{readv,writev}` for the
  operation, plus `fentry/mm_access` with a `kprobe/mm_access` fallback for target
  identity — the common permission gate, where the target is an already-resolved
  `task_struct`. eBPF-only.
- Prior art: Tracee, Datadog CWS and Falco all pair a syscall hook with a kernel-function
  hook when the syscall argument is not what they need; this follows that pattern. Falco
  implements both syscalls fully and ships them disabled by default.
- Semantics/state: emitted on exit with the raw return; entry/exit correlated through
  bounded per-thread state with mandatory cleanup.
- Event/flag: `QUARK_EV_PROCESS_VM_ACCESS`, `QQ_PROCESS_VM_ACCESS`
- Coverage: both syscalls, both directions; same-process calls suppressed; a call naming
  a reaped pid never reaches `mm_access` and is not reported; a namespaced caller now
  resolves correctly. `/proc/<pid>/mem` is an accepted gap, better modelled on file open.
- Compatibility: 4.18+, dual fentry/kprobe with runtime fallback; a failed attach means
  the event is absent, never a failed probe load.
- Volume: one event per syscall, not per iovec. No dedup, deliberately — see Proof.

## Proof

- Build `make CLANG=clang-17` PASS; `t_process_vm_access @ ebpf ... ok` covers a
  successful read and write, a suppressed same-process call, a reaped-pid call, and
  `target_start_time_ns != 0`.
- Cold Fleet acceptance `20260910T102925Z-process-vm-access` (Endpoint/Agent 9.4.2).
- Persona volume gate `20260910T114814Z-process-vm-access-volume` on accepted build
  `endpoint-9.4.2-aba0979dc8d75bb9` (PASS, 1h): **0** `process_vm_access` documents
  against 91,643 documents/hour across six personas. That is why there is no dedup.

Captured during acceptance:

```json
{"operation": "write", "target_pid": 5738, "target_start_time_ns": 2736128582547,
 "local_iovcnt": 1, "remote_iovcnt": 1, "remote_addr": 138511200903168,
 "bytes_requested": 8, "bytes_transferred": 8}
```

## Known limitation: RHEL 9.3

`t_process_vm_access` is excluded on the RHEL 9.3 shard, alongside the `t_memfd` and
`t_shmget` exclusions already there. On `5.14.0-362.8.1.el9_3` the syscall-enter
tracepoint record does not match the hard-coded offsets this file uses to read syscall
arguments, so `remote_iovcnt`, `remote_addr` and `bytes_requested` are wrong on that
kernel — observed as `assert(qpva->remote_iovcnt == 1)` failing for a call made with
`riovcnt == 1`, while `target_pid` and `target_start_time_ns` (read from the task at
`mm_access`) were correct on the same event.

This is pre-existing and not specific to this feature: `memfd_create`, `shmget` and
`process_vm_{readv,writev}` are the only probes reading tracepoint arguments this way,
and RHEL 9.3 is the only 9.x shard affected. The durable fix is to read arguments
through BTF's `struct syscall_trace_enter` so libbpf relocates the offsets per kernel,
which repairs all three at once; that is #417 rather than this PR, since it needs
verification on that exact kernel build.

## Companion PRs

- endpoint-package: https://github.com/elastic/endpoint-package/pull/787
- endpoint-dev: https://github.com/elastic/endpoint-dev/pull/21420


Aegrah:
CI triage for [build 1423](https://buildkite.com/elastic/quark/builds/1423), since none of
the three red jobs are this branch's code and the one fix in it is still unproven.

**All three failures are `apt-get` (exit 100).**

- `build amd64 in docker` — the image build's `apt-get install bison bpftool clang ...`
  failed, so no `initramfs.gz` was produced. Every distro shard then reported
  `WAITING_FAILED` rather than running, because they `depends_on: make_docker`. **That
  includes rhel 9.3, so the exclusion added here has not actually been exercised yet.**
- `valgrind amd64` and `test-go` — `E: Unable to locate package linux-tools-6.11.0-26-generic`.

The `linux-tools` one looks agent-dependent rather than branch-dependent. Build 1421
(`nicholasberlin/perf-record-*`, 2026-09-09) passed the same step on an agent running
kernel `7.0.0-14`, installing `linux-tools-7.0.0-14-generic` without trouble. This branch's
builds landed on agents running `6.11.0-26`, whose `linux-tools` package is no longer in
the Ubuntu archive as a superseded ABI. Same pipeline, different agent image.

Separately, from build 1422 (before the docker step started failing): fedora 43 and 44
failed every eBPF test with `quark_queue_open: Operation not supported`, including
`t_probe`, which opens a queue with default flags and so never loads this feature's
programs. Those shards had moved from kernel 7.1.13 to **7.2.4** since `main` last built
on 2026-09-08. That looks like it affects `main` equally and is worth a maintainer's eye
independently of this PR.

nicholasberlin:
> Known limitation: RHEL 9.3

Fixed here: https://github.com/elastic/quark/pull/422

Aegrah:
Superseded by the stack https://github.com/elastic/quark/pull/427 → https://github.com/elastic/quark/pull/428 → https://github.com/elastic/quark/pull/429 (private probes → public event and tests → Go bindings). The redesign keeps the hybrid architecture (syscall result + `mm_access` target identity) but replaces the named syscall tracepoints with ABI-aware raw entry/exit hooks (ia32/x32 compat calls were invisible), keeps attempts rejected before target lookup, makes the iovec fields honest bounded snapshots instead of a first-vector "address/size" (a zero-length decoy first vector and a partial transfer inside one vector were both reproduced), fixes the release-branch timestamp/ring-loss accounting, adds the optional-source retry in both Endpoint watchers, and replaces the weak tests (namespace/TID identity, compat entry, permission denial, partial transfer, map pressure, cleanup). Gates: preflight `20260916-pva-v2-release-final` PASS, Fleet acceptance `20260916T172021Z-process-vm-access` PASS (accepted build `endpoint-9.4.2-5564d77d41a8132d`); the persona volume gate was explicitly waived by the owner for this round and is recorded as waived, not passed. Per the workflow rule a redesign opens new PRs instead of force-pushing a different design here; closing this one.

PR #418, opened 2026-09-14
Add file_access events for opens of watched names

Aegrah:
Part 1 of 4 of the `file_access` stack, bottom-up:

1. **this PR** — core event: probe, anchors, in-kernel dedup, `quark-test`, `quark-mon -a` (experimental)
2. https://github.com/elastic/quark/pull/419 — failed opens (`EACCES`/`EPERM`/`ENOENT`)
3. https://github.com/elastic/quark/pull/420 — target task of `/proc/<pid>/` opens

## Summary

Adds `QUARK_EV_FILE_ACCESS`: an event for a *completed* `open(2)` of a file whose name the
consumer registered, so reads of ssh keys, cloud credentials, `/etc/shadow`, shell history and
`/proc/<pid>/{mem,maps,...}` become visible. The existing file events cannot carry this:
`CREATE`/`REMOVE`/`MOVE` never fire for a plain read and `MODIFY` is content only.

Issue: https://github.com/elastic/endpoint-dev/issues/19153 (procfs part: https://github.com/elastic/endpoint-dev/issues/21205)

- Hook/backend: `do_filp_open` return — the same point the create event already uses — as
  `fexit` with the existing `kprobe`+`kretprobe` fallback (arguments carried by a new entry
  kprobe, `EBPF_EVENTS_STATE_FILP_OPEN`). Below the syscall demux, so `open`/`openat`/
  `openat2`/`creat`/`execve` (`FMODE_EXEC`)/io_uring/compat all pass through it. eBPF-only.
- Prior art: Tracee hooks `security_file_open` (blind to DAC-denied opens), Falco the syscall
  exit (result but an unresolved path and no in-kernel path filter), Datadog CWS entry +
  `vfs_open` + `terminate_walk` + exit with leaf/parent name approvers before resolving,
  Tetragon `security_file_permission` per read. Followed Datadog's model (completed-open
  semantics, name approvers before resolution, write = write-intent open) and Tracee's field
  set; diverged on the hook because quark already sits on `do_filp_open`'s return for the
  create event — one hook sees the final file and the requested flags, and PR 2 reuses it for
  the error result.
- Semantics/state: emitted on success with the resolved path, inode, mode, uid/gid, size,
  requested `open_flags` and `fmode`. Nothing is resolved before a cheap anchor test: the leaf
  dentry name and up to three ancestor names (`FILE_ACCESS_ANCESTORS`) are looked up in a
  userspace-populated hash map (`quark_queue_file_access_name_add(3)`); the consumer applies
  the real path patterns to the resolved path, the map only has to be a superset.
- Dedup (the volume lever): once per process life, file and access class (`O_PATH` > `exec` >
  write intent, i.e. `O_WRONLY`/`O_RDWR`/`O_CREAT`/`O_TRUNC`/`O_APPEND` > `read`), keyed
  `(tgid, dev, ino)` with `(start_time, self_exec_id)` as the identity so a stale entry never
  suppresses; every failure mode costs a duplicate, never a lost event. The class is claimed
  **after** `ebpf_ringbuf_write()` succeeded (`ebpf_dedup__test()` before the event is built,
  `ebpf_dedup__set()` after it is out), so an event lost to a full ring buffer leaves the next
  open of that file reportable instead of silencing it for the process life. The key/value
  and helpers are shared with the mprotect probe (`elastic-ebpf/GPL/Events/Dedup.h`); mprotect
  keeps claiming before output through the unchanged `ebpf_dedup__test_and_set()`, now the
  composition of the two, so its behaviour on `main` is untouched (retrying there is a
  separate follow-up).
  RHEL 8 keeps `self_exec_id` in `task_struct_rh` (partial type in `vmlinux_extra.h`).
- Event/flag: `QUARK_EV_FILE_ACCESS`, `QQ_FILE_ACCESS` (`1<<16`, marked experimental in
  `quark.h`). #416 (`process_vm_access`) uses `QQ 1<<16`/`EBPF 1<<22` on its base — whichever
  of the two merges second renumbers.
- Compatibility: 4.18+. Off by default: rodata gate the verifier prunes, maps `max_entries 0`
  with `bpf_map__set_autocreate(false)` unless `QQ_FILE_ACCESS` is set, entry kprobes autoloaded
  only with the flag — a queue without the flag pays nothing. The file-event side of the shared
  hook has its own gate (`file_events_enabled`, from `QQ_FILE`), so a queue with
  `QQ_FILE_ACCESS` alone gets no `QUARK_EV_FILE`.
- Verifier on 4.18.0-553 (RHEL 8.10, CI toolchain), `fexit__do_filp_open` processed
  instructions, limit 1,000,000: stock `main` with `QQ_FILE` 644,517; stack head with `QQ_FILE`
  only 644,554 (the gates); the whole stack with `QQ_FILE_ACCESS` only 483,118, with
  `QQ_FILE|QQ_FILE_ACCESS` — what Endpoint loads — **913,173 (91.3 %)**, the number to watch when
  anything else is added to that hook (round 1 measured 886,844; the +26,329 is the second
  dedup lookup on the emit paths and PR 2's base-directory read before the dedup test).
  PR 2's requested-string branch is the expensive half.
- Volume: one event per (process life, file, class). On the persona volume gate the whole
  feature (this PR + PR 2 + PR 3, default sensitive-path list) produced 3,217 `open` documents/hour across six personas (budget 12,000) and 91,480 total
  documents/hour (budget 158,000) on the pre-review code; the gate on this lineage is running
  and its numbers replace these.

- Files opened under a hardlink or bind mount whose name matches no anchor are invisible (same
  limit Tetragon documents); `open_by_handle_at` bypasses `do_filp_open` (needs
  `CAP_DAC_READ_SEARCH`).
- A reader that inherits or receives an fd (`SCM_RIGHTS`) is not reported; the opener is.
- Parent anchors reach three directory levels below the anchored directory; deeper files are
  not seen. Compile-time bound, reviewed here, deliberately not a policy knob (every open pays
  for the walk).
- kretprobe kernels (4.18) can miss returns under concurrency (`maxactive`), as the existing
  create event already does; fentry kernels are unaffected.
- io_uring opens on 5.5–5.11 are attributed to the io-wq kernel thread (Falco/Tracee share this).

- Opens **through** a procfs magic link (`/proc/<pid>/{exe,cwd,root,fd/<n>,map_files/<range>,
  ns/*}`) are opens of the link's target: path walk follows the link before `do_filp_open`
  returns, so the dentry seen is the target's, under its own parents. Reported only if the
  target itself matches an anchor, and then as a plain file. (The first review round wrongly
  called `map_files/<range>` reachable through a `map_files` parent anchor; see #420 for the
  procfs side and the test.)

- `file_events_enabled` gate: `QQ_FILE_ACCESS` without `QQ_FILE` no longer leaks create
  events; `quark-test` errors on any `QUARK_EV_FILE` seen without `QQ_FILE`.
- Dedup key/value/`test_and_set` hoisted to `Dedup.h` and used by mprotect as well.
- `quark_queue_file_access_name_reset(3)` documented as `int`; kretprobe-fallback comment
  explains the `op == NULL` flag source; `FILE_ACCESS_ANCESTORS` comment; `O_PATH` class test.

### Changed in review round 2 (`c3ed084`)

- The dedup class is claimed only once the event is out (`ebpf_dedup__test` / `ebpf_dedup__set`
  split, see Dedup above). Test: fill the ring buffer with undrained opens under the parent
  anchor until `quark_queue_stats.lost` moves, read a file while it is full (`lost` moves
  again), drain until idle past `hold_time`, and the next read of that file must be reported.
- Valgrind: the trailing `quark_queue_file_access_name_reset()` smoke check is guarded with
  `if (!in_valgrind)` like `t_trusted_pid` (libbpf's `bpf_map_get_next_key()` memsets
  `bpf_attr` short of `flags` and takes a NULL first key). The ring-buffer flood is skipped
  under valgrind too. No suppressions file, no zeroed-key half-fix.
- `quark_queue_get_event(3)`: an event lost to a full ring buffer does not count as reported.

## Proof

- Build: `make CLANG=clang-17` PASS; CI toolchain (`make docker`) loads every program on
  4.18.0-553.
- Sensor test: `t_file_access @ ebpf ... ok` (Ubuntu 22.04, 6.8.0-1066-gcp) — read/write/exec/`O_PATH`
  classes, repeat open not re-emitted, new process re-emits, unanchored name absent, nested
  create three levels under a parent anchor reported and four levels not, no `QUARK_EV_FILE`
  without `QQ_FILE`, a read lost to a full ring buffer is reported on the next open;
  `t_mprotect` still ok with the shared dedup helper.
- Valgrind (`make test-valgrind`, the CI step that was red): guard in place; the Buildkite valgrind step on this push is the check. A local `make test-valgrind` on the GCP builder is not representative: its kernel/libbpf setup floods valgrind with `unhandled eBPF command 36` noise on the stock tests too (2.4 M reports) and aborts, so it proves nothing either way.
- Preflight `20260915T094418Z-sensitive-file-access-preflight` PASS (quark build + full `quark-test`, package, schema handoff, Endpoint build, gtests, EAF). Cold Fleet acceptance and the 1 h persona volume gate on this exact lineage are **running**; their run IDs replace this line when they finish (the pre-review code passed `20260914T105843Z-sensitive-file-access` / `20260914T120054Z-sensitive-file-access-volume`). The gates ran on the
  release-aligned branch `sensitive-file-access-9.4.2` (`5f97a6c`, same code on `ba3b465`).

```text
  FACC	access=R open_flags=0100000 error=0
  FACC	path=/tmp/lsw-sfa/id_rsa
  FACC	mode=0100600 uid=1033 gid=1034 size=7 inode=2328882 fmode=0x494a801d
```

## Companion PRs


Aegrah:
Review fix-ups pushed as `7b6d5dd` (fast-forward, history kept): `file_events_enabled` rodata gate so `QQ_FILE_ACCESS` alone gets no `QUARK_EV_FILE` (+ quark-test check), dedup key/value/`test_and_set` hoisted to `elastic-ebpf/GPL/Events/Dedup.h` and used by mprotect too, `.Ft int` for `name_reset`, kretprobe-fallback flag comment, `O_PATH` class test, `FILE_ACCESS_ANCESTORS` comment. Body updated with the re-measured 4.18 verifier numbers (stack head 886,844 with `QQ_FILE|QQ_FILE_ACCESS`; `QQ_FILE` only unchanged at 644,554 vs 644,517 on main). `map_files` removed from the accepted gaps: a parent anchor reaches it.

nicholasberlin:
Re-reviewed `7b6d5dd`. All the review items are addressed and the fix-ups are correct: the `file_events_enabled` gate is set only under `QQ_FILE` and cuts the create/memfd/shmem branch before it runs; `Dedup.h` is line-for-line the mprotect logic (compared against `main`), and the 4.18 types, `.Ft int`, fallback comment, `O_PATH` test and ancestor comment are all in. One thing left, and it is what has kept Buildkite red on every commit of the stack while `main` is green:

**Valgrind step** (build 1444, "build and run quark-test under valgrind amd64"). `t_file_access @ ebpf` prints `ok`, but valgrind records 8 errors from 2 contexts, all under `quark_queue_file_access_name_reset (bpf_queue.c:1847)` -> `bpf_map_get_next_key (bpf.c:509)` -> `syscall`:

```
Syscall param bpf(attr->flags) points to uninitialised byte(s)
Syscall param bpf(attr->key) points to unaddressable byte(s)
  Address 0x0 is not stack'd, malloc'd or (recently) free'd
```

libbpf memsets `bpf_attr` only up to `offsetofend(next_key)` and passes `NULL` as the first key; valgrind flags both. 7 `get_next_key` calls (6 anchors + the terminating `ENOENT`) + 1 `NULL`-key report = 8. `quark_queue_trusted_pid_reset()` has the same loop and `t_trusted_pid` skips itself under valgrind for exactly this ("short memset before sys_bpf"); `t_file_access` calls the reset at `quark-test.c:1633` with no guard.

The `t_tty` death right after (`quark-test: tty child didn't exit cleanly`, run stops at 24/49) is fallout, not a second bug: with `--error-exitcode=1` a forked child inherits valgrind's error count, so the tty child's `exit(0)` becomes exit 1 and the parent `errx`es. On `main` nothing is recorded, so the child exits 0.

The reset is a trailing smoke check with nothing asserted after it, so guarding just that line keeps the rest of the test under valgrind, where its value is (event parsing, the `quark_file_access` allocation, the #419 path join):

```c
	/*
	 * Same libbpf false positive as t_trusted_pid: bpf_map_get_next_key()
	 * memsets bpf_attr short of flags and takes a NULL first key.
	 */
	if (!in_valgrind)
		assert(!quark_queue_file_access_name_reset(&qq));
```

A zeroed key instead of `NULL` in `name_reset()` would only remove the second report, the first is inside libbpf; a suppressions file for both resets is the right long-term fix but not for this PR. Please also add the valgrind step to the Proof section. With that in, this is an approve from me; #419-#421 pick it up on the merge-forward and need nothing of their own.

Aegrah:
Round-2 fix-ups pushed as `c3ed084` (fast-forward): the dedup class is now claimed only after `ebpf_ringbuf_write()` returned 0 (`ebpf_dedup__test()` / `ebpf_dedup__set()` split in `Dedup.h`; `test_and_set()` kept as their composition for mprotect, whose behaviour is unchanged), with a test that fills the ring buffer, loses a read and asserts the next read is reported; and the `quark_queue_file_access_name_reset()` smoke check is guarded with `if (!in_valgrind)` as proposed. `t_file_access` + `t_mprotect` ok on Ubuntu 22.04 (6.8) for this head and the stack head. The Buildkite valgrind step on this push is the check for the guard — a local `make test-valgrind` on my GCP builder is not representative (its kernel/libbpf floods valgrind with `unhandled eBPF command` noise on stock tests and aborts). RHEL 8 verifier numbers for the round-2 heads follow in a comment once re-measured; the body carries the pre-round-2 figures marked as such.

Aegrah:
RHEL 8 verifier numbers for the round-2 heads, `4.18.0-553.159.1.el8_10`, CI toolchain (`make docker NO_GO=y`), every program loads; `fexit__do_filp_open` processed instructions, limit 1,000,000:

| load | round 1 (`1b121db`) | round 2 (`c45fc14`) |
|---|---|---|
| `QQ_FILE` only | 644,554 | 644,554 |
| `QQ_FILE_ACCESS` only | 457,217 | 483,118 |
| `QQ_FILE\|QQ_FILE_ACCESS` (what Endpoint loads) | 886,844 (88.7 %) | **913,173 (91.3 %)** |

The +26k is the second map lookup in `ebpf_dedup__set()` on both emit paths plus the base-directory read moved ahead of the failure dedup test. Still under the limit with ~87k headroom; bodies of #418–#420 updated with these figures. (The static `quark-mon` still segfaults in NSS on the RHEL 8 userland after loading, as noted before — verifier log only.)

PR #419, opened 2026-09-14
Report failed opens of watched names in file_access

Aegrah:
Part 2 of 4 of the `file_access` stack, bottom-up:

1. https://github.com/elastic/quark/pull/418 — core event: probe, anchors, in-kernel dedup, `quark-test`, `quark-mon -a`
2. **this PR** — failed opens (`EACCES`/`EPERM`/`ENOENT`)
3. https://github.com/elastic/quark/pull/420 — target task of `/proc/<pid>/` opens

Based on `file-access-core`; the diff shown here is this branch's commits only.

## Summary

A denied read of a credential file is the detection case, so opens of a watched name that
fail with `EACCES`, `EPERM` or `ENOENT` are now reported too, from the same `do_filp_open`
return. There is no dentry on that branch: the requested pathname is read into per-cpu
scratch (≤256 bytes), its leaf and up to three parent components are matched against the
anchor map, and the base directory of a relative name is resolved from the `dfd` or cwd so
the consumer can rebuild the path the caller meant.

Issue: https://github.com/elastic/endpoint-dev/issues/19153

- Semantics: `quark_file_access` gains `requested`, `base_dir`, `error`, `dfd` and flags
  `QUARK_FILE_ACCESS_F_FAILED` / `QUARK_FILE_ACCESS_F_RELATIVE`; `path` is the join of
  `base_dir` and `requested` when the base is known. Kretprobe kernels take the arguments
  from the entry state the success path already carries.
- Errno set: Datadog CWS emits `EACCES`/`EPERM` by default and `ENOENT` only opt-in; `ENOENT`
  is in by design here (a probe for a credential file that does not exist yet is signal)
  and is what made in-kernel dedup of failures necessary.
- Dedup: once per process life, **base directory**, errno, requested name (32-bit FNV-1a of
  the string) **and access class** — the same class bitmap as the file map, in a second LRU
  (`elastic_ebpf_file_access_fail_seen`) with the same `(start_time, self_exec_id)` contract
  and the same claim-after-output order as #418. The base directory is the `(dev, ino)` of
  the directory a relative name was resolved against (`fs->pwd` or the dirfd's `f_path`),
  resolved **before** the dedup test — the event needed it anyway for `base_dir` — and 0/0
  for an absolute name; the dirfd *number* is not in the key (`AT_FDCWD` after a `chdir()`
  and a reused descriptor name different directories). An `ENOENT` read probe followed by an
  `ENOENT` create attempt on the same name are two events, as on the success path; the same
  relative name missing from two directories is two events. A hash collision costs a
  suppressed failure of a different name for that process life, nothing else.
- Verifier: the hash/leaf scan is a 256-iteration bounded loop with the running state kept in
  map memory (a tracked register made the verifier re-verify the tail per loop exit). On
  4.18.0-553 (CI toolchain) the whole stack takes `fexit__do_filp_open` to 483,118
  processed instructions with `QQ_FILE_ACCESS` only and 913,173 (91.3 %) with
  `QQ_FILE|QQ_FILE_ACCESS` (see #418); this branch's requested-string scan is the expensive
  half.
- Volume: counted in the same persona volume gate as PR 1 (3,217 `open` documents/hour across six personas on the pre-review code, successes and
  failures together; this lineage's gate is running).

- The failure path is the **requested string**, not a resolved path: `/etc/../etc/shadow`, a
  symlink, or a `/proc/self/fd/N` magic link whose spelled name carries no anchor is not
  reported (confirmed on a VM: DAC-denied `cat /tmp/not-shadow-2 -> /etc/shadow` produces
  nothing). The successful `O_PATH` open and the symlink's creation are the observable
  precursors. Same limit as Datadog's `open.syscall.path` and Falco's `fd.name`; closing it
  needs the resolved dentry on the denial path (`may_open`/`do_open` are static), not worth the
  cost. Textual `..`/`.` canonicalisation in quark userspace is a possible follow-up.
- A relative failed name only exposes the components it spells out to the parent-anchor test.
- Requested names longer than 256 bytes are not reported. A leaf of 64 bytes or more is not
  looked up as a leaf anchor, but its parents still are.
- kretprobe kernels (4.18) can miss returns under concurrency, as in PR 1.

- Failure dedup is per access class (was a single "failed" bit; the comment claiming the
  same contract as the file map was untrue).
- A too-long leaf no longer short-circuits the parent-anchor test on the failure path.
- Manpage: the failed-open matching rule sits with `quark_queue_file_access_name_add(3)`.

### Changed in review round 2 (`c85aff4`)

- The failure key carries the base directory's `(dev, ino)` instead of the dirfd number
  (deterministic aliasing across `chdir()` / reused descriptors: a scanner probing `id_rsa`
  across directories was reported once). Key 16 -> 24 bytes.
- The class is claimed only once the event is out, as on the success path (#418).
- `quark_queue_get_event(3)`: once per requested name, base directory, error and access class.
- Picks up #418's valgrind guard on the merge-forward.

## Proof

- Build: `make CLANG=clang-17` PASS; `make docker` loads on 4.18.0-553.
- Sensor test: `t_file_access @ ebpf ... ok` — `ENOENT` on a watched leaf (absolute, and
  relative with `F_RELATIVE` + `base_dir`), an unwatched failed name absent, a same-class
  repeat failure not re-emitted, a different-class failure on the same name re-emitted,
  failed absolute and relative opens under a parent anchor, a 71-byte leaf under a parent
  anchor reported, `EACCES` on an `execve` of a non-executable watched name (`__FMODE_EXEC`
  in `open_flags`), the same relative name missing from two directories through `chdir()` and
  through a reused dirfd number reported four times with the right `base_dir` and `dfd` each
  while the same-directory repeat stays suppressed.
- Gates (whole feature): preflight `20260915T094418Z-sensitive-file-access-preflight` PASS; cold Fleet acceptance and the 1 h persona volume gate on this lineage are **running**, IDs to follow (pre-review code: `20260914T105843Z-sensitive-file-access` / `20260914T120054Z-sensitive-file-access-volume` PASS).

```text
  FACC	access=R open_flags=01100000 error=13
  FACC	path=/etc/sudoers.d/buildkite_agent
  FACC	requested=/etc/sudoers.d/buildkite_agent dfd=-100
```

## Companion PRs


Aegrah:
Review fix-ups pushed as `21f3f99` (fast-forward): failed-open dedup now per access class (`file_access_class(open_flags)` as the bit, `FILE_ACCESS_CLASS_FAILED` gone, comment fixed), a too-long leaf no longer skips the parent-anchor test on the failure path, manpage sentence moved next to `name_add`. quark-test: same-class repeat suppressed, other-class repeat reported, 71-byte leaf under a parent anchor reported.

Aegrah:
Round-2 fix-up pushed as `c85aff4` (fast-forward, on the merge-forward of #418): the failure key is `(tgid, errno, base dev, base inode, hash)` — the base directory is resolved before the dedup test and the dirfd number is gone — and the class is claimed after output like the success path. Test: the same relative name missing from two directories through `chdir()` and through a reused dirfd number yields four events with the right `base_dir`/`dfd`; the same-directory repeat stays suppressed. One thing worth knowing from the first attempt: `BPF_CORE_READ(scratch->path.dentry, ...)` makes libbpf emit a CO-RE relocation against `struct file_access_scratch`, which has no kernel counterpart, and the program is rejected at load; the dentry pointer is loaded into a local first.

Aegrah:
Buildkite on #419-#421 was red on one assertion in the valgrind job (	_file_access, the exec EACCES case): valgrind's execve wrapper opens the target read-only to inspect it, then checks it is executable and fails the syscall itself, so the kernel never sees the execve and the child produced a completed read of 
oexec instead of the failed exec. Test-only fix-up 9551f96 skips that case under valgrind (the rest of the test still runs there); merged forward to #420/#421. Verified on Ubuntu 22.04 with and without VALGRIND=1.

PR #420, opened 2026-09-14
Name the target task of /proc/<pid>/ opens in file_access

Aegrah:
Part 3 of 4 of the `file_access` stack, bottom-up:

1. https://github.com/elastic/quark/pull/418 — core event: probe, anchors, in-kernel dedup, `quark-test`, `quark-mon -a`
2. https://github.com/elastic/quark/pull/419 — failed opens (`EACCES`/`EPERM`/`ENOENT`)
3. **this PR** — target task of `/proc/<pid>/` opens

Based on `file-access-failed`; the diff shown here is this branch's commits only.

## Summary

An open of another process's `mem`, `maps`, `environ` or `syscall` is a process access, and
the detection needs to know which process. Entries under `/proc/<pid>/` carry their task in
`PROC_I(inode)->pid`, so the probe reads the target tid, tgid and start time from the inode
regardless of how the caller spelled the pid (`self`, a tid, a pid inside a pid namespace)
and flags the event `QUARK_FILE_ACCESS_F_PROCFS`. The pre-4.19 `pid_links`/`pids` layout
(RHEL 8) is handled through a partial type in `vmlinux_extra.h`.

Issue: https://github.com/elastic/endpoint-dev/issues/21205

- Fields: `target_pid`, `target_tid`, `target_start_time` on `struct quark_file_access`;
  `quark-mon` prints them on the `FACC` line. `target_start_time` is on the same clock as
  `quark_process.proc_time_boot`, so `quark_process_lookup(target_pid)->proc_time_boot ==
  target_start_time` identifies the target across pid reuse (scraped processes carry jiffies
  precision, documented).
- `F_PROCFS` marks **any** entry whose inode has a pid, including the caller's own
  (`/proc/self/maps`): the flag says "target fields are meaningful". A target that already
  exited still yields `target_tid`; `target_pid` and `target_start_time` are then 0.
- **`QUARK_FILE_ACCESS_NAME_OTHER_TASK`** (new anchor role, `quark-mon -a @name`): a name
  registered with it only reports `/proc/<pid>/` entries that belong to *another* thread
  group than the opener; the check is one compare on the target already read from the
  inode, done **before** the dedup test so a skipped self-read never claims a slot. This is
  how the consumer keeps glibc's `pthread_getattr_np` read of its own `maps`, or a runtime
  reading its own `status`/`cmdline`, out of the stream while still seeing the same read of
  another process. Endpoint registers every `/proc` pattern with it except `mem`, `maps` and
  `environ`, whose self-reads stay visible (self-injection through `/proc/self/mem`,
  introspection).
- Dedup: folded into the file map — `/proc/<pid>/maps` has its own inode per (pid, entry), so
  `(tgid, dev, ino)` already carries the target identity. procfs inode churn costs a duplicate,
  never a loss.
- Cost: one `PROC_I` read on anchored opens whose superblock is procfs; nothing on other opens.
  Verifier on 4.18.0-553 (CI toolchain): the whole stack is 483,118 processed
  instructions with `QQ_FILE_ACCESS` only and 913,173 (91.3 %) with `QQ_FILE|QQ_FILE_ACCESS`
  (see #418).

- Failed procfs opens (DAC `EACCES` on another uid's `mem`, `ESRCH` for a dead pid) have no
  inode; the consumer only has the caller-namespace-relative `<pid>` spelled in the requested
  string (PR 2) and no kernel-resolved target. `OTHER_TASK` cannot apply there either.

- **Magic links are opens of their target.** `exe`, `cwd`, `root`, `fd/<n>`,
  `map_files/<range>`, `ns/<name>` are symlink inodes (`proc_pid_get_link`,
  `proc_map_files_get_link`); an ordinary open follows them before `do_filp_open` returns, so
  the dentry the probe sees is the target's, under the target's parents, on the target's
  filesystem: no `map_files` parent anchor match, no `F_PROCFS`, no target task. Such an open
  is reported only if the target itself matches an anchor, as a plain file. Only an
  `O_PATH|O_NOFOLLOW` open of the link itself is a procfs entry (and it carries the pid like
  the others). Keeping the procfs identity across the link would need a hook on the path walk
  with per-thread state on every open, for entries #21205 does not ask for (`mem`, `maps`,
  `environ`, ... are regular procfs files) — excluded and documented in
  `quark_queue_get_event(3)`, owner decision. (The first review round wrongly called
  `map_files/<range>` reachable through a parent anchor; the first draft called it
  unanchorable.)

- The "open option" (self-read handling) is decided and implemented as the `OTHER_TASK`
  anchor role above; no `file.Ext.procfs.self` field, no unconditional kernel skip.
- `file_access_anchored()` returns the matched roles; procfs target resolution moved into
  per-cpu scratch ahead of the dedup test.
- `quark-mon -a` syntax: `name` (leaf), `name/` (directory), `@name` (other tasks only);
  manpage wording ("a task's entry"), `target_start_time` clock documented;
  `pid_link___4_18`/`task_struct___4_18` live in `vmlinux_extra.h` from this PR on.

### Changed in review round 2 (`dd20777`)

- The `map_files` claim is withdrawn; magic-link traversal is an accepted, documented gap
  (above, `quark_queue_get_event(3)`, `quark.h`). Test: the child's first `map_files` entry
  opened `O_RDONLY` under a `map_files/` parent anchor is not reported, `O_PATH|O_NOFOLLOW` is,
  with the child as target.
- Picks up #418 (claim-after-output, valgrind guard) and #419 (base-directory failure key)
  on the merge-forward; the failed procfs path claims after output too.

## Proof

- Build: `make CLANG=clang-17` PASS; `make docker` loads on 4.18.0-553 (4.18 layout).
- Sensor test: `t_file_access @ ebpf ... ok` — under `maps` (plain) and `@status`
  (`OTHER_TASK`): our own `/proc/self/maps` is reported with ourselves as the target, our own
  `/proc/self/status` is not, a child's `status` and `maps` carry `F_PROCFS`,
  `target_pid == target_tid == child`, `target_start_time > 0`; registering `OTHER_TASK`
  alone (no `LEAF`/`PARENT`) is rejected; a `map_files/<range>` entry opened `O_RDONLY` is
  not reported and `O_PATH|O_NOFOLLOW` is, with `F_PROCFS` and the child as target.
- Gates (whole feature): preflight `20260915T094418Z-sensitive-file-access-preflight` PASS; cold Fleet acceptance and the 1 h persona volume gate on this lineage are **running**, IDs to follow (pre-review code: `20260914T105843Z-sensitive-file-access` / `20260914T120054Z-sensitive-file-access-volume` PASS). EAF `test_file_access_events_linux` asserts the Endpoint side of `OTHER_TASK`
  (`/proc/self/status` absent, `/proc/<child>/status` present with the target fields).

```text
  FACC	access=R open_flags=0100000 error=0
  FACC	path=/proc/5471/maps
  FACC	mode=0100444 uid=0 gid=0 size=0 inode=61264 fmode=0x494a801d
  FACC	target_pid=5471 target_tid=5471 target_start_time=1857452343941
```

## Companion PRs


Aegrah:
Review fix-ups pushed as `1b121db` (fast-forward). The open option is decided (owner: keep `mem`, `maps` and `environ` self-reads visible): new anchor role `QUARK_FILE_ACCESS_NAME_OTHER_TASK` (`quark-mon -a @name`) drops opens of the caller's own `/proc/<pid>/` entries for the names that carry it; `file_access_anchored()` returns the matched roles and the procfs target is resolved into scratch before the dedup test so a skipped self-read never claims a slot. Endpoint registers every `/proc` pattern except `mem`/`maps`/`environ` with it. Also: `target_start_time` documented as the `proc_time_boot` clock, 'a task's entry' wording, `vmlinux_extra.h` 4.18 types move into this PR. quark-test: own `maps` reported, own `status` (under `@status`) not, child's `status`/`maps` name the child, `OTHER_TASK` alone rejected.

Aegrah:
Round-2 pushed as `dd20777` (fast-forward). Owner decision: exclude and document. The `map_files` claim is withdrawn — you are right that an ordinary open follows the magic link before `do_filp_open()` returns and the dentry seen is the mapped file's — and the accepted gap is stated generally (`exe`, `cwd`, `root`, `fd/<n>`, `map_files/<range>`, `ns/*`) in `quark_queue_get_event(3)`, `quark.h` and the body. Test: the child's first `map_files` entry opened `O_RDONLY` under a `map_files/` parent anchor is not reported, `O_PATH|O_NOFOLLOW` is, with `F_PROCFS` and the child as target. Picks up #418/#419 on the merge-forward.

PR #423, opened 2026-09-14
Never suppress security events for trusted pids

nicholasberlin:
## Motivation

A pid in the trusted map is meant to shed the volume of a chatty but benign application. The check also gated `module_load`, `ptrace`, `shmget` and `memfd_create`, which are rare, high signal events that a trusted application produces at a negligible rate.

A published bypass abuses exactly that: a root process inserts its own tgid into the trusted map, loads a kernel module, and removes the entry. The `module_load` event never reaches userspace. See https://matheuzsecurity.github.io/hacking/elastic-trusted-pid-bypass/

## Change

- Drop `ebpf_events_is_trusted_pid()` from the four security probes in `Process/Probe.bpf.c`.
- Document in `State.h` which classes trust is allowed to suppress: file, network and mprotect.
- `CHANGES` entry, since consumers relying on trusted pids to hide these classes see a behaviour change.

Nothing changes for callers that never add a trusted pid. `mprotect` stays gated: it is deduplicated per process life but a JIT heavy trusted application still produces a bounded burst of them.

## Test

`t_trusted_pid` used to only exercise add and reset. It now trusts the test process itself, checks that a loopback connect produces no socket event while a `shmget` still does, then resets trust and checks the socket events come back so the suppression half is not vacuous.

Sockets are used instead of files because the RHEL 8.5 to 8.7 CI kernels lack file events. Rebased on #422, which fixed `shmget` on RHEL 9.3, so no exclusion is needed there.

## Verification

krun VMs, `quark-test -v t_trusted_pid`:

| Kernel | Result |
| --- | --- |
| Fedora 43, 7.2.5 | ok |
| RHEL 8.5, 4.18.0-348 | ok |
| RHEL 9.3, 5.14.0-362 | ok (after rebase on #422) |

Negative: with only the probe change reverted, the test fails on Fedora 43 with `drain_for_pid: Timer expired` waiting for the `shmget` event, so it pins this change.

Full eBPF suite (`quark-test -b`) passes on Fedora 43.

Not exercised in the suite: `ptrace` and `module_load` under a trusted pid. Both are the same one line removal as `shmget`.


christos68k:
@nicholasberlin Ping me if this is ready for review (it's still marked draft)

nicholasberlin:
This is working as designed. We are trusting the pids in the map there should be no events from them. PRs to mitigate and event on the published bypass are https://github.com/elastic/quark/pull/424, https://github.com/elastic/quark/pull/425, and https://github.com/elastic/quark/pull/426.

PR #424, opened 2026-09-14
Lock down the eBPF maps against bpf(2) strangers

nicholasberlin:
## Motivation

The trusted pids map is written from userspace and only ever read by the probes, yet nothing stopped a BPF program from writing to it. A privileged process can still update the map through `bpf(2)`, but with `BPF_F_RDONLY_PROG` it can no longer smuggle a writer in as BPF code: the verifier refuses any program that writes the map, ours or a stranger's.

Every other map except the ring buffer is the opposite case: written only by the probes, never by userspace. Freezing them with `BPF_MAP_FREEZE` means no process can plant or erase quark's state through `bpf(2)`. Without that a stranger could pre-mark its own mprotect transitions as seen in `mprotect_seen`, or poison `sk_to_tgid` to misattribute connections.

This does not address the `bpf(2)` update route to the trusted map itself, which is the subject of #425.

## Change

`bpf_queue.c`:

- Probe once with a throwaway hash map for `BPF_F_RDONLY_PROG` and `BPF_MAP_FREEZE`. Both arrived in Linux 5.2 and older kernels reject them outright. On such kernels the maps are left as they were, with a debug log line.
- `BPF_F_RDONLY_PROG` on `elastic_ebpf_events_trusted_pids` and on `elastic_ebpf_events_init_buffer`, which is only ever copied from.
- After load, `bpf_map_freeze()` on every map except the trusted map and the ring buffer. libbpf's internal maps (`.rodata`) are skipped, libbpf freezes those itself. Maps never created because autocreate is off are skipped by that flag, since libbpf still hands them a placeholder fd.
- Only `EINVAL` from the probe means an old kernel; anything else is warned about. The memlock bump now runs before the probe so a tight `RLIMIT_MEMLOCK` cannot silently disable the hardening.

`CHANGES` entry.

## Test

`t_trusted_map_rdonly` finds the trusted map the way a stranger would, by walking every map id and matching type, sizes, entry count and name prefix (the kernel truncates names to 15 characters, so every `elastic_ebpf_events_*` map shares one), then checks:

- the flag is set on the live map
- userspace writes still work, both the library calls and a raw `bpf(2)` update and delete
- a hand assembled program calling `map_lookup_elem` on the map loads, while the same program calling `map_update_elem` is refused with `EACCES`

`t_map_freeze` checks a raw `bpf(2)` update and delete on `sk_to_tgid` fail with `EPERM`, and that the init buffer both refuses a write and carries `BPF_F_RDONLY_PROG`.

Both skip on kernels without the features.

## Verification

krun VMs, `quark-test -v t_trusted_pid t_trusted_map_rdonly t_map_freeze`:

| Kernel | Result |
| --- | --- |
| Fedora 43, 7.2.5 | ok |
| RHEL 8.5, 4.18.0-348 | ok (both backported, not skipped) |
| RHEL 9.3, 5.14.0-362 | ok |

Negative: with only the trusted map flag reverted, `t_trusted_map_rdonly` aborts on the `map_flags` assertion. With only the freeze loop skipped, `t_map_freeze` aborts on the `EPERM` assertion. Each test pins its half.

Full eBPF suite (`quark-test -b`) passes on Fedora 43.


christos68k:
@nicholasberlin Ping me if this is ready for review (it's still marked draft)

nicholasberlin:
> Ping me if this is ready for review (it's still marked draft)

This is ready for review @christos68k (thanks)

bpf_queue.c
```diff
@@ -1440,6 +1502,30 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
...
 
+	/*
+	 * The trusted map is written from userspace and only ever read by the
+	 * probes. Mark it read-only for programs so the verifier refuses any
+	 * BPF writer, ours or a stranger's; a root process can still update it
+	 * through bpf(2), but can't smuggle a writer in as BPF code. The init
+	 * buffer is only ever copied from, by anyone, so it gets the flag too
+	 * and is frozen below.
+	 */
+	harden = bpf_map_harden_supported();
```
stanek-michal, line 1513: comment said we want to bump memlock first, then check with probe, but here it's reversed?

bpf_queue.c
```diff
@@ -1469,6 +1555,13 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
 		goto fail;
 	}
 
+	if (harden & BPF_HARDEN_FREEZE) {
+		if (bpf_maps_freeze(p) == -1)
+			goto fail;
```
christos68k, line 1560: Isn't it better to continue with a warning here given that we're operating under best-effort?

Also another (maybe far-fetched) scenario could be that kernels with backported functionality are allowing`BPF_MAP_FREEZE` on `BPF_MAP_TYPE_HASH` (which is what we test) but fail on different type maps.

bpf_queue.c
```diff
@@ -1152,9 +1153,70 @@ open_probes(void)
...
 
+/*
+ * Two ways to take a map out of a stranger's hands, both since Linux 5.2:
+ * BPF_F_RDONLY_PROG makes the verifier reject any program that writes to it,
+ * BPF_MAP_FREEZE rejects any further write from userspace, through any fd.
+ * Older kernels reject the flag and the command themselves, so probe with a
+ * throwaway map and report what the kernel takes.
+ */
+#define BPF_HARDEN_RDONLY_PROG	(1 << 0)
+#define BPF_HARDEN_FREEZE	(1 << 1)
```
christos68k, line 1164: Should we move these to `quark.h`?

bpf_queue.c
```diff
@@ -1152,9 +1153,70 @@ open_probes(void)
...
+	bpf_object__for_each_map(m, p->obj) {
+		if (m == p->maps.elastic_ebpf_events_trusted_pids ||
+		    m == p->maps.ringbuf || bpf_map__is_internal(m) ||
+		    !bpf_map__autocreate(m))
+			continue;
+		if ((fd = bpf_map__fd(m)) < 0)
+			continue;
+		if (bpf_map_freeze(fd) != 0) {
+			qwarn("bpf_map_freeze %s", bpf_map__name(m));
+			return (-1);
```
christos68k, line 1209: Same rationale here, we should probably continue with a warning and not abort on first failure

```suggestion
```

bpf_queue.c
nicholasberlin: Good catch. Force pushed a fix, here's the [compare link](https://github.com/elastic/quark/compare/a0b547dde1ff0ebc9ff8d4180342737505a9c40a..d21177e1c895610183625c15d06363ec65ddf5e0)

bpf_queue.c
nicholasberlin: I'd rather keep these private. They are the return bits of one static helper in bpf_queue.c and nothing outside the file reads them, while quark.h is the installed public header.

bpf_queue.c
nicholasberlin: Agreed, switched to a warning in the force push ([compare link](https://github.com/elastic/quark/compare/a0b547dde1ff0ebc9ff8d4180342737505a9c40a..d21177e1c895610183625c15d06363ec65ddf5e0)).

bpf_queue.c
nicholasberlin: Agreed, switched to a warning in the force push ([compare link](https://github.com/elastic/quark/compare/a0b547dde1ff0ebc9ff8d4180342737505a9c40a..d21177e1c895610183625c15d06363ec65ddf5e0)).

bpf_queue.c
```diff
@@ -1443,10 +1513,35 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
...
+	 * through bpf(2), but can't smuggle a writer in as BPF code. The init
+	 * buffer is only ever copied from, by anyone, so it gets the flag too
+	 * and is frozen below.
+	 */
+	harden = bpf_map_harden_supported();
+	if (harden & BPF_HARDEN_RDONLY_PROG) {
+		if (bpf_map__set_map_flags(p->maps.elastic_ebpf_events_trusted_pids,
+		    BPF_F_RDONLY_PROG) != 0) {
+			qwarn("bpf_map__set_map_flags trusted_pids");
+			goto fail;
```
christos68k:
```suggestion
```

bpf_queue.c
```diff
@@ -1443,10 +1513,35 @@ bpf_queue_open1(struct quark_queue *qq, int use_fentry)
...
+		}
+		if (bpf_map__set_map_flags(p->maps.elastic_ebpf_events_init_buffer,
+		    BPF_F_RDONLY_PROG) != 0) {
+			qwarn("bpf_map__set_map_flags init_buffer");
+			goto fail;
```
christos68k:
```suggestion
```

christos68k (dismissed):

bpf_queue.c
nicholasberlin: This function will only fail if the map is already created:
https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L10847-L10853
which would be a logic bug we should avoid by bailing.

Agreed?

christos68k: Correct, took it as a generic failure on setting the flag.

nicholasberlin:
@christos68k and @stanek-michal thanks for the reviews, the follow-on PRs revealed a bug in the test code. [Some stack garbage needing clearing](https://github.com/elastic/quark/compare/d21177e1c895610183625c15d06363ec65ddf5e0..5e02bd74e2a99bd6a2b3b93fa701d215e9cde018).

christos68k (approved):

PR #428, opened 2026-09-16
Expose process_vm_access attempt telemetry

Aegrah:
Part 2 of 3, bottom-up: [private probes](https://github.com/elastic/quark/pull/427) → [public event and tests](https://github.com/elastic/quark/pull/428) → [Go bindings](https://github.com/elastic/quark/pull/429).

## Summary

Expose one process_vm_access attempt per observed syscall exit, including signed result, flags and explicit target/snapshot validity. Capacities are present only for complete checked snapshots of at most eight vectors. The first remote vector is labeled as a snapshot, not a transfer range or total. Unresolved attempts retain the original signed PID and caller PID namespace.

Add four regression tests covering failures, partial transfer, namespace/TID identity, compat entry, concurrency, map pressure and cleanup. Public event/queue bit 17 avoids mprotect and pending file-access allocations. Quark-mon, loss statistics and documentation describe the contract.

Issue: https://github.com/elastic/endpoint-dev/issues/21204

## Validation

Preflight: `20260916-pva-v2-release-final`. Fleet acceptance: `20260916T172021Z-process-vm-access`, accepted build `endpoint-9.4.2-5564d77d41a8132d` (Stack/Agent/Endpoint 9.4.2). Main integration was validated separately. Full persona-volume gate explicitly waived by the user; no production-volume claim.

Native x64/ARM64, forced eBPF kprobe attachment, RHEL 8 and exact Rocky 9.3 pass. A small enabled/disabled getpid benchmark measured ~53 ns extra per syscall on the shared builder; this does not establish general production overhead.

```text
local_snapshot_status=0 remote_snapshot_status=0 local_capacity=8 remote_capacity=8
```

Companions: https://github.com/elastic/endpoint-package/pull/800 · https://github.com/elastic/endpoint-dev/pull/22027


Aegrah:
Buildkite was red on the valgrind job only. Reproduced on a builder: valgrind refuses the nonleader-thread target's `clone(CLONE_VM | CLONE_SIGHAND | CLONE_THREAD)` (`Unsupported clone() flags: 0x10900`), so the child died before the handshake; and its syscall checker counts the deliberately invalid local vectors (unaddressable `lvec`, `iovcnt` beyond the array, `~0` length) as errors before the kernel can reject them. Fix-up `3c18711` (fast-forward): the thread target uses the pthread-shaped flag set that `clone_start()` already uses, and under valgrind the test skips the compat-ABI cases, those three argument-validation scenarios, and the identity/state/concurrent tests (pid namespaces, credential drops, map pressure, concurrent children) the way `t_trusted_pid` does. The plain read/write/failure scenarios still run under valgrind, where the event parsing is. Plain run: all four tests pass on 6.8; merged forward to #429.

PR #432, opened 2026-09-18
Verify the file emit paths once, not per call site

nicholasberlin:
## Summary

The `do_filp_open()` return hook verifies at **644k** processed instructions on RHEL 8's 4.18 with `QQ_FILE` (64 % of the 1M limit), and #418 adds another emit path to the same program (86 % alone, 91 % at the head of that stack). This PR brings the hook down to **~209k** and `fentry__vfs_rename` from 337k to 168k, without changing any event, so the file_access stack lands with headroom instead of a number to watch.

## Where the instructions come from

The verifier counts the instructions it *walks* while simulating every path, not the size of the program. To attribute the hook's 626k (`main`, `QQ_FILE`, `4.18.0-553`, laptop clang 18), I replaced one piece at a time with a stub that returns an empty string and reloaded; the drop from stock is that piece's cost:

| piece of `fexit__do_filp_open` | cost (processed insns) | share |
|---|---|---|
| dentry path resolver, walked three times | ~505k | 81 % |
| cgroup (kernfs) path resolver, walked three times | ~117k | 19 % |
| everything else: pid/cred/ns fills, file info, memfd and tmpfs checks | ~4.6k | < 1 % |
| **total** | **626k** | |

One emit through `prepare_and_send_file_event()` is ~209k, the same as every single-emit file program (`vfs_write`, `chmod`, `unlink` all sit at 208k). The hook pays it **three times** because the function is a **static** subprogram reached from the create, memfd and shmem branches, and the verifier re-walks a static function from every call site. Deleting two of the three branches brings the hook to 209k, one third. `vfs_rename__enter()` pays twice, one resolver call per name.

## The change

- `prepare_and_send_file_event()` and a new `vfs_rename__resolve_path()` become **global BPF functions** (non-static, `__noinline`). The verifier checks a global function once, on its own, and treats the call as opaque at every site.
- The oldest verifier we load on (RHEL 8's 4.18) only accepts scalar arguments and returns for global functions and rejects a pointer taken from the probe context (`Caller passes invalid args`). So the `struct file *` and the dentries travel as `u64` via `ebpf_ptr_to_scalar()`, a `bpf_probe_read_kernel()` of the pointer's own stack slot; a cast alone doesn't change the verifier's view of a register. Every access to those pointers already went through `BPF_CORE_READ`, which is a probe read and takes any value.
- The rename resolver always writes `new_path`; picking the buffer by argument would be two verifier states again. The caller copies the first result to `old_path` (one 4 KiB probe read per tracked rename).
- The `/dev/shm` prefix filter moves inside the emit function as a flag, same reason.
- The laundering read runs only on the emit paths; the miss path of the hook, every other open on the host, is unchanged.
- **Preemption** (the first CI run): verifiers from 6.10, where `bpf_preempt_disable()` appeared, to 6.14, before the sleepability analysis that lifts the rule, reject `global function calls are not allowed with preemption disabled`; RHEL 9.8's 5.14 carries the backport. That is why Fedora 39/40, RHEL 9 and Ubuntu 25.04 failed while 4.18, 5.11, 6.12 and 7.x passed. The two global functions now disable preemption themselves, around their per-cpu event buffer and resolver scratch, and the six wrappers that reach them (`do_filp_open`/`do_file_open` exits, `vfs_rename` entries) no longer do. Everything else on those paths is task-keyed state (`ebpf_events_state`, the rename scratch space), not per-cpu.

Merging the three call sites into one call of the static function is **not** enough: measured 419k, because the shmem path carries a different prefix pointer and the verifier keeps it as a separate state.

## Numbers

`4.18.0-553`, CI toolchain (`make docker NO_GO=y`, clang 21), processed instructions with `QQ_FILE`:

| program | main | this PR |
|---|---|---|
| `fexit__do_filp_open` | 644,526 | **215,071** |
| `fentry__vfs_rename` | 362,315 | **181,229** |

No other program changes. For #418 that means roughly 430k with `QQ_FILE|QQ_FILE_ACCESS` instead of 860k. (The laptop's clang 18 measures ~3 % lower on both columns; the attribution above used it.)

## Proof

- `make CLANG=clang-18` and `make docker NO_GO=y` build; every program loads on `4.18.0-553`, `4.18.0-477` (RHEL 8.8), `5.11.22` (Fedora 32, the oldest bpf kernel in CI) and `7.2.5` (Fedora 44).
- `quark-test` in qemu, laptop build, full suite 49/49 ok on: RHEL 8.10 (`4.18.0-553`), RHEL 9 (`5.14.0-687`), RHEL 10 (`6.12.0-211`), Fedora 39 (`6.11.9`), Fedora 40 (`6.14.5`), Fedora 44 (`7.2.5`), Ubuntu 24.04/26.04 (`7.0.0-31`), Ubuntu 25.04 (`6.14.0-37`). File tests ok on RHEL 8.8 (`4.18.0-477`) and Fedora 32 (`5.11.22`, oldest bpf kernel in CI). CI toolchain build: file tests ok on `4.18.0-553`.
- Negative check: with the `old_path` copy removed from `vfs_rename__enter()`, `t_file_rename` fails on `qf->old_path != NULL`; with the probe change reverted and the original test assumption (create and move in one event) it fails on `op_mask`, which is how the aggregation note in the test came about.
- Rename had no test. `t_file_rename` creates a file, renames it, and checks `op_mask & QUARK_FILE_OP_MOVE`, `path`, `old_path` and the inode on the aggregated event. Excluded from the RHEL 8.5-8.7 "no file" jobs like the other file tests.

## Follow-ups (not here)

- `sched_process_exec` (378k) and `fentry__vfs_rename`'s remaining 168k are single walks and stay as they are; a cheaper resolver or a smaller `PATH_RESOLVER_MAX_COMPONENTS` is the lever for those.
- #418 conflicts trivially in `do_filp_open__exit()` (it inserts a call and two gates above the branch this PR touches); its `file_access__open()` is a single-site static subprogram and needs nothing.


gogochan (approved):

nicholasberlin:
## How to reproduce the instruction counts locally

The numbers in the description come from loading the real object on a RHEL 8 kernel under qemu and reading libbpf's debug log. No root needed, only `/dev/kvm` access (it falls back to TCG without it, just slower), plus `qemu-system-x86_64`, `lynx`, `rpm2cpio`, `cpio` and docker.

**1. Build both trees with the CI toolchain.** Clang version changes the counts by a few percent, so use the docker image to compare against the table above. `NO_GO=y` skips the Go binaries, which the initramfs doesn't need.

```sh
git worktree add ../quark-main main
(cd ../quark-main && make docker NO_GO=y)
make docker NO_GO=y          # in the PR checkout
```

**2. Save this helper** as `insns.sh`. It boots the kernel, loads quark-mon with `QQ_FILE`, prints one line per BPF program with its `processed N insns` count and exits once the backend is up. `LIBBPF_LOG_LEVEL=debug` is passed on the kernel command line, which init turns into an environment variable, so libbpf prints the verifier summary for each successful load.

```sh
#!/bin/bash
# usage: insns.sh <quark tree> [rhelver]    e.g. insns.sh ../quark-main 8
cd "$1" || exit 1
timeout -s KILL 600 ./krun-rhel.sh initramfs.gz "${2:-8}" \
    quark-mon -F -vv -m 1 -s LIBBPF_LOG_LEVEL=debug 2>&1 |
awk '
	/-- BEGIN PROG LOAD LOG --/ { match($0, /prog \x27[^\x27]+\x27/); prog = substr($0, RSTART+6, RLENGTH-7) }
	/processed [0-9]+ insns/     { for (i=1;i<=NF;i++) if ($i=="processed") printf "%-40s %10s\n", prog, $(i+1) }
	/using .* for backend/       { exit }
' | sort -k2 -n -r
```

**3. Run and diff.** `8` fetches the latest Rocky 8 kernel-core rpm (`4.18.0-553` today), the oldest verifier we support and the one closest to the 1M limit. Each run downloads the rpm (~60 MB) and takes a couple of minutes.

```sh
./insns.sh ../quark-main > insns-main.txt
./insns.sh .             > insns-pr.txt
diff -y -W 110 insns-main.txt insns-pr.txt
```

**Expected on `4.18.0-553`** (my run today, CI clang):

| program | main | this PR |
|---|---|---|
| `fexit__do_filp_open` | 644,517 | 215,073 |
| `fentry__vfs_rename` | 362,315 | 181,234 |

All other programs are identical between the two columns. The global functions don't show up as separate lines: the verifier checks them inside the load of the program that calls them, so the per-program total already includes one walk of each.

To try another kernel pass its version as the second argument, e.g. `8.8`, `9`, `10`; `krun-fedora.sh` and `krun-ubuntu.sh` take the same shape. quark-mon never exits on its own in the VM, so if you drop the `awk`, kill qemu by pid when you're done.

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -332,17 +340,22 @@ static void prepare_and_send_file_event(struct file *f,
...
-                ebpf_ringbuf_write(&ringbuf, event, EVENT_SIZE(event), 0);
-        }
+    // skip event if only /dev/shm files are wanted and the path does not start with it
+    if (devshm_only) {
+        if (size >= (long)sizeof(DEVSHM_STRING) - 1 &&
+            is_equal_prefix(field->data, DEVSHM_STRING, sizeof(DEVSHM_STRING) - 1))
+            ebpf_ringbuf_write(&ringbuf, event, EVENT_SIZE(event), 0);
     } else {
         ebpf_ringbuf_write(&ringbuf, event, EVENT_SIZE(event), 0);
     }
```
biscout42: would it do the same?
```suggestion
    if (!devshm_only || (size >= (long)sizeof(DEVSHM_STRING) - 1 &&
            is_equal_prefix(field->data, DEVSHM_STRING, sizeof(DEVSHM_STRING) - 1))) {
            ebpf_ringbuf_write(&ringbuf, event, EVENT_SIZE(event), 0);
}
```

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -288,15 +288,23 @@ int BPF_KPROBE(kprobe__vfs_unlink)
...
-                                        const char *path_prefix,
-                                        int path_prefix_len)
+// if devshm_only is set the event is only sent if the file path starts with /dev/shm
+//
+// Global function, see ebpf_ptr_to_scalar(): do_filp_open__exit() reaches
+// this from three places (create, memfd, shmem) and the verifier would walk
+// the path resolvers once per place otherwise. `file` is the struct file *
+// as a scalar. Disables preemption itself for the per-cpu event buffer and
+// resolver scratch: the caller must not, a global function may not be called
+// with preemption disabled on 6.10-6.14 verifiers.
```
christos68k: Clarifying why we can't pass `struct *file` or `void*` here.


```suggestion
// as a scalar to appease verifier. Disables preemption for the per-cpu event buffer and
// resolver scratch: the caller must not as a global function may not be called
// with preemption disabled on 6.10-6.14 verifiers.
```

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -288,15 +288,23 @@ int BPF_KPROBE(kprobe__vfs_unlink)
...
+__noinline int prepare_and_send_file_event(u64 file, u32 type, u32 devshm_only)
```
christos68k: We can pass `devshm_only` as `bool` and `type` as `enum ebpf_event_type` (verifier should have no issue with this) for better clarity on the human reader side:

```suggestion
__noinline int prepare_and_send_file_event(u64 file, enum ebpf_event_type type, bool devshm_only)
```

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -288,15 +288,23 @@ int BPF_KPROBE(kprobe__vfs_unlink)
 }
 
 // prepare a file event and send it to ringbuf.
-// if path_prefix is non-NULL then event will only be sent to ringbuf if file path has that prefix
-static void prepare_and_send_file_event(struct file *f,
-                                        enum ebpf_event_type type,
-                                        const char *path_prefix,
-                                        int path_prefix_len)
+// if devshm_only is set the event is only sent if the file path starts with /dev/shm
```
christos68k: If we switch to `bool`:

```suggestion
// if devshm_only the event is only sent if the file path starts with /dev/shm
```

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -594,14 +615,12 @@ static int vfs_rename__enter(struct dentry *old_dentry, struct dentry *new_dentr
...
-    struct path p;
-    p.mnt    = state->rename.mnt;
-    p.dentry = old_dentry;
-    ebpf_resolve_path_to_string(ss->rename.old_path, &p, task);
-    p.dentry = new_dentry;
-    ebpf_resolve_path_to_string(ss->rename.new_path, &p, task);
+    // mnt was read back from the state map, so the verifier already sees a
+    // scalar; the dentries come from the probe context and need the detour.
+    u64 mnt = (u64)state->rename.mnt;
+    vfs_rename__resolve_path(mnt, ebpf_ptr_to_scalar(old_dentry));
```
christos68k:
```suggestion
    if (vfs_rename__resolve_path(mnt, ebpf_ptr_to_scalar(old_dentry)))
        goto out;
```

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -594,14 +615,12 @@ static int vfs_rename__enter(struct dentry *old_dentry, struct dentry *new_dentr
...
+    bpf_probe_read_kernel(ss->rename.old_path, PATH_MAX, ss->rename.new_path);
+    vfs_rename__resolve_path(mnt, ebpf_ptr_to_scalar(new_dentry));
```
christos68k:
```suggestion
    if (vfs_rename__resolve_path(mnt, ebpf_ptr_to_scalar(new_dentry)))
        goto out;
```

christos68k (dismissed):
LGTM

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/c52960a6f748d9bb7b218ef8516e231967a6af3c

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/c52960a6f748d9bb7b218ef8516e231967a6af3c

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/c52960a6f748d9bb7b218ef8516e231967a6af3c

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/c52960a6f748d9bb7b218ef8516e231967a6af3c

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/c52960a6f748d9bb7b218ef8516e231967a6af3c

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: yep, same. https://github.com/elastic/quark/pull/432/commits/68ee6b9d9d6bf01e109aaec92a252b396af85975

biscout42 (dismissed):

christos68k (dismissed):

biscout42 (dismissed):

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -693,9 +679,42 @@ int BPF_KRETPROBE(kretprobe__filename_renameat2)
 
     return r;
 }
```
christos68k, line 681:
```suggestion
}

```

elastic-ebpf/GPL/Events/File/Probe.bpf.c
```diff
@@ -693,9 +679,42 @@ int BPF_KRETPROBE(kretprobe__filename_renameat2)
...
+    ebpf_resolve_path_to_string(ss->rename.new_path, &p, task);
+    r = 0;
+
+out:
+    preempt_enable();
+    return r;
+}
+
+// Runs with preemption enabled, see vfs_rename__resolve_path(); the state and
+// scratch maps are keyed by task, only the resolver's scratch is per-cpu.
```
christos68k:
```suggestion
```

christos68k (dismissed):

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/d3d049e657f11507fcd9560c402c1cbd8b2314e1

elastic-ebpf/GPL/Events/File/Probe.bpf.c
nicholasberlin: https://github.com/elastic/quark/pull/432/commits/d3d049e657f11507fcd9560c402c1cbd8b2314e1

christos68k (approved):

